PR #12, opened 2020-02-05
Feature/apply style

mradkov:


thepiwo:
please add `fontawesome` as npm dependency. If only needed for one icon, consider removing it

thepiwo:
mixins/variables probably won't need to be imported multiple times

thepiwo:
if the aeternity spinner resource is no longer needed we can remove it, as well as the logo

thepiwo:
Overall the whole content looks really small, barely readable on a big screen

thepiwo:
The whole explanation header as well as reference to the tipping extension is missing with this updated design, is this intentional?

thepiwo:
- Links are not clickable
 - Adresses look clickable but are not
 - Comment and Heart Icon do nothing

thepiwo:
- Comment and Heart Icon look really pixalated and won't scale on higher dpi screens.
 - search box has no margin to other header

PR #16, opened 2020-02-12
Added functionality for retrieving link previews

stanislav-slavov:


stanislav-slavov:
> looks good, will displaying be in a separate PR?

The visualization would be in the open PR for Corona new until we make it compatible.

PR #18, opened 2020-02-19
Add fiat value

mradkov:


src/utils/currency.js
```diff
@@ -0,0 +1,29 @@
+const wrapTry = async (f) => {
```
thepiwo: there is already a wrapTry defined in backend, no need to do it twice

src/views/TipsList.vue
```diff
@@ -200,6 +233,17 @@
           }
         }
 
+        const currencyInstance = new Currency();
+        const getCurrencyRates = currencyInstance.getRates().catch(console.error);
+        const currencyRates = await Promise.resolve(getCurrencyRates);
```
thepiwo: why not inline await in previous line?

src/views/TipsList.vue
```diff
@@ -134,6 +148,25 @@
...
+        if(languageChoose == 'cn'){
+          this.isChineseActive = true;
+          this.defaultCurrency = 'cny';
+        }else{
+          this.isChineseActive = false;
+          this.defaultCurrency = 'eur';
+        }
+        this.displayCurrency();
+      },
+      isPreviewToBeVisualized(tip){
```
thepiwo: why add preview function in this PR?

src/views/TipsList.vue
thepiwo: I think the whole currency thins should EITHER be not blocking OR parallel with fetching tips

PR #19, opened 2020-02-19
Added language filtering of tips

stanislav-slavov:


src/views/TipsList.vue
```diff
@@ -226,6 +233,14 @@
           }else{
             this.tips = tips;
           }
+          if(typeof langTips !== 'undefined' && langTips.length > 0){
```
thepiwo: seems overlapping with the logic few lines aboves, can we unify this?

PR #27, opened 2020-03-02
fixed sorting function code duplication

thepiwo:


src/App.vue
```diff
@@ -25,31 +25,28 @@
...
     computed: {
-      ...mapGetters(['tips', 'current', 'account', 'isLoggedIn']),
+      ...mapGetters(['tips', 'tipSortBy', 'current', 'account', 'isLoggedIn']),
     },
     methods: {
-      sort(sorting) {
-        this.sorting = sorting;
-
-        switch (this.sorting) {
+      sort() {
```
kenodressel: maybe move this function to the setTipsOrdering action

src/App.vue
```diff
@@ -159,11 +156,17 @@
       }
     },
     async created() {
-      await this.reloadData(true);
+      this.$store.subscribe((mutation, _) => {
```
kenodressel: this can be avoided when the sorting is in the actions

PR #29, opened 2020-03-02
Show loading on index page

stanislav-slavov:
closes https://github.com/aeternity/aepp-raendom/issues/53

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -2,7 +2,7 @@
   <div id="app" class="min-h-screen">
     <div class="content min-h-screen max-w-desktop">
       <div class="min-h-screen wrapper" ref="wrapper">
-        <router-view></router-view>
+        <router-view :showLoading="showLoading"></router-view>
       </div>
     </div>
   </div>
--- a/src/App.vue
+++ b/src/App.vue
@@ -2,7 +2,7 @@
   <div id="app" class="min-h-screen">
     <div class="content min-h-screen max-w-desktop">
       <div class="min-h-screen wrapper" ref="wrapper">
-        <router-view :showLoading="showLoading"></router-view>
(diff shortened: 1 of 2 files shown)


src/App.vue
```diff
@@ -2,7 +2,7 @@
   <div id="app" class="min-h-screen">
     <div class="content min-h-screen max-w-desktop">
       <div class="min-h-screen wrapper" ref="wrapper">
-        <router-view></router-view>
+        <router-view :showLoading="showLoading"></router-view>
```
thepiwo: I think we should't give props to router-view

thepiwo: @kenodressel what do you think=

kenodressel: I think it should be solved differently. You could, for example, only render the router-view if the loading is done, if the data is so vital to the application. Otherwise I would recommend handling the loading state in the individual components which are depending on the data.

src/views/TipsList.vue
```diff
@@ -150,14 +150,6 @@
...
       openExplorer(address) {
         return this.explorerUrl + address
-      },
-      showLoadingTips() {
-        let load = setInterval(() => {
-          if (this.tips.length > 0){
-            this.showLoading = false;
-            clearInterval(load);
-          }
-        }, 200);
```
davidyuk: It could be just a computed variable:
```js
showLoadingTips() {
  return !!(this.tips && this.tips.length);
}
```

thepiwo:
@stanislav-slavov I think this should be https://github.com/aeternity/aepp-raendom/issues/53

thepiwo:
I'll take over work on this :)

thepiwo:
closing this for now

PR #30, opened 2020-03-02
cleanup app

thepiwo:


src/App.vue
```diff
@@ -85,58 +73,35 @@
...
+            tip.fiatValue = (tip.total_amount * rates.aeternity[this.current.currency]).toFixed(2);
+            return tip;
+          })//.filter(tip => tip.total_amount * (rates.aeternity['usd']).toFixed(2) > 0.01);
+        }
 
         stats.height = await aeternity.client.height();
-        this.$store.commit('UPDATE_STATS', stats);
 
-        this.asyncAddCurrency();
+        this.$store.dispatch('setTipsOrdering', tipOrdering);
```
kenodressel: Should we check if these are valid before we push them to the store?

thepiwo: I don't think so for now, but we should reconsider if they are needed in the store at all.

PR #77, opened 2020-03-05
Refactor store

thepiwo:
closes https://github.com/aeternity/aepp-raendom/issues/24

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -30,7 +30,7 @@
       }
     },
     computed: {
-      ...mapGetters(['current', 'tipSortBy']),
+      ...mapGetters(['settings', 'tipSortBy']),
     },
     methods: {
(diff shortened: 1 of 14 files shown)


src/App.vue
```diff
@@ -30,9 +30,10 @@
       }
     },
     computed: {
-      ...mapGetters(['current', 'tipSortBy']),
+      ...mapGetters(['settings', 'tipSortBy']),
     },
     methods: {
+      ...mapActions(['setLoggedInAccount', 'setTipsOrdering', 'updateTips', 'updateTopics', 'updateStats', 'updateCurrencyRates', 'setTipSortBy']),
```
kenodressel: not sure about this. I personally like it better when its an explicit store communication.

thepiwo: I think this is good practice :D

PR #79, opened 2020-03-05
Issue 74: Removed url from tip record

stanislav-slavov:


src/components/tipRecords/TipRecordComponent.vue
```diff
@@ -14,9 +14,6 @@
           <tip-title :tip="tip" :goToTip="goToTip"></tip-title>
         </div>
       </div>
-      <div>
```
thepiwo: as this is removed, whats linking to the url now?

src/components/tipRecords/TipRecordComponent.vue
```diff
@@ -31,7 +31,9 @@
         </div>
       </div>
     </div>
-    <div class="tip__article col-4" v-if="isPreviewToBeVisualized(tip)">
+    <div class="tip__article col-4 position-relative" v-if="isPreviewToBeVisualized(tip)">
```
thepiwo: `isPreviewToBeVisualized` also within this?

stanislav-slavov: These are minor changes depending on one another. That's why I combined it.

thepiwo: but the if isPreviewToBeVisualized is also nested inside which doesn't make sense to me

stanislav-slavov: I caught your point, the section should be hidden if there are no results. Only one check will stay

PR #87, opened 2020-03-05
Issue #57 Issue/remove component string

stanislav-slavov:


thepiwo:
@stanislav-slavov can your resolve the conflicts?

src/views/CreateProfile.vue
```diff
@@ -256,16 +256,16 @@
...
+  import RightSection from '../components/layout/RightSection.vue';
 
   export default {
     name: 'CreateProfile',
     components: {
-      'header-component': HeaderComponent,
-      'left-section': LeftSectionComponent,
-      'right-section': RightSectionComponent,
+      'custom-header': Header,
+      'left-section': LeftSection,
```
thepiwo: the mapping is not needed for all if the name is the same

stanislav-slavov: There is a <header> html5 tag, so probably we have to leave it for the header.

PR #90, opened 2020-03-06
Feature/download link browser detection

mradkov:
- detect browser
    - different download url based on browser
    - TODO: update chrome url when published
- update AEX2 name to `Superhero`

fixes: https://github.com/aeternity/aepp-raendom/issues/44

### DIFF
--- a/src/components/layout/RightSection.vue
+++ b/src/components/layout/RightSection.vue
@@ -7,7 +7,7 @@
           Wallet
         </div>
         <div class="section__body clearfix">
-          <div class="side__button"><a href="//github.com/superherowallet/wallet/releases/latest">Install Wallet</a></div>
+          <div class="side__button"><a :href="downloadUrl">Install Wallet</a></div>
         </div>
       </div>
       <div class="section trending">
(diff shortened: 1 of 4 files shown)


src/components/layout/RightSection.vue
```diff
@@ -55,10 +56,26 @@
...
     data() {
-      return {}
+      return { 
+        browser: detect()
+      }
     },
     computed: {
       ...mapGetters(['topics', 'isLoggedIn']),
+      downloadUrl () {
+        switch(this.browser && this.browser.name) {
```
thepiwo: this shouldn't work, switch on a boolean?

mradkov: It works actually, but if you are not happy with this - I can change the implementation.

kenodressel: just wrap it in an if(this.browser) :)

src/components/layout/RightSection.vue
```diff
@@ -7,7 +7,7 @@
           Wallet
         </div>
         <div class="section__body clearfix">
-          <div class="side__button"><a href="//github.com/superherowallet/wallet/releases/latest">Install Wallet</a></div>
+          <div class="side__button"><a :href="downloadUrl">Install Wallet</a></div>
```
thepiwo: it should probably open in a new tab

PR #95, opened 2020-03-06
#94 Issue/adapt colors sizes

stanislav-slavov:


src/App.vue
```diff
@@ -131,7 +131,7 @@
 <style>
 
   html {
-    font-size: 135%;
+    font-size: 125%;
```
thepiwo: why not 100%?

stanislav-slavov: 1rem on 100% is equal to 16px. 125% is 20px - It is alot easier to calculate.

src/components/layout/LeftSection.vue
```diff
@@ -41,32 +44,32 @@
             {{stats.total_tips_length}}
           </div>
           <div class="overview__label">
-            Tips
+            Tips Aggregated
           </div>
         </div>
-        <div class="overview__item" v-if="stats.comments">
+        <!-- <div class="overview__item" v-if="stats.comments">
```
thepiwo: why is the stats that were required by yani removed?

src/views/TipsList.vue
```diff
@@ -171,9 +165,14 @@
...
   }
   .container.wrapper{
+    padding-top: 0;
     min-height: 4rem;
   }
    .actions__container{
+      top: 0;
+      position: sticky;
+      position: -webkit-sticky;
+      z-index: 10;
```
thepiwo: why z-index?

src/views/TipsList.vue
```diff
@@ -171,9 +165,14 @@
     }
   }
   .container.wrapper{
+    padding-top: 0;
     min-height: 4rem;
   }
    .actions__container{
+      top: 0;
+      position: sticky;
+      position: -webkit-sticky;
```
thepiwo: if not supported widely or with polyfil we shouldn't use it

src/views/UserProfileView.vue
```diff
@@ -268,14 +265,20 @@
...
-    .address{
-      color: $standard_font_color;
-    }
-    .count{
-      color: $light_font_color;
-      font-size: .6rem;
+    top: 0;
+    position: sticky;
+    position: -webkit-sticky;
+    z-index: 10;
```
thepiwo: why z-index?

src/components/layout/LeftSection.vue
```diff
@@ -118,38 +121,46 @@
...
 
-      .navigation__item, .side__button {
-        font-size: .75rem;
+      .navigation__item {
+        font-size: 1rem;
         text-transform: capitalize;
-        margin-bottom: 0.7rem;
-
+        margin-bottom: 1rem;
+        a a{
```
thepiwo: nested a doesn't seem to make sense

src/views/UserProfileView.vue
stanislav-slavov: Because Position sticky is fixed when scrolled and it goes under the content.

PR #106, opened 2020-03-09
Issue/remove address trimming

stanislav-slavov:


stanislav-slavov:
I will reopen it after a consultation with the designers.

PR #123, opened 2020-03-11
Resolved profile header overlapping and tip note visualization

stanislav-slavov:


src/views/UserProfileView.vue
```diff
@@ -262,7 +262,7 @@
 
   .profile__header{
     top: 0;
-    z-index: 10;
```
thepiwo: is there a solution without z-index?

stanislav-slavov: The icons have higher z-index(they are clickable and have popups) than the sticky section. It's a simple and only solution in this case.

thepiwo: The Icons shouldn't have a z-index as well, as I said, is there a solution WITHOUT z-index? (cc @kenodressel )

stanislav-slavov: Screenshot 2020-03-12 at 13 32 27

stanislav-slavov: I could remove the z-index from the icons but still the z-index: 10; should stay for the header, because the profile section is with relative position(z-index is unset) and below the header => the header will go under it without specified z-index.

PR #124, opened 2020-03-12
Build using @vue/cli

### LINKED ISSUE #42: Use vue/cli as a boilerplate
They have great hot reloading out of the box, it should increase development speed.

davidyuk:
fixes #42


thepiwo:
@davidyuk whats the difference between this and https://github.com/kenodressel/aepp-boilerplate/pull/7?

tests/e2e/specs/test.js
```diff
@@ -0,0 +1,8 @@
+// https://docs.cypress.io/api/introduction/api.html
+
+describe('My First Test', () => {
```
thepiwo: should probably be adjusted before merge

tests/unit/example.spec.js
```diff
@@ -0,0 +1,12 @@
+import { shallowMount } from '@vue/test-utils';
+import HelloWorld from '@/components/HelloWorld.vue';
+
+describe('HelloWorld.vue', () => {
```
thepiwo: should probably be adjusted before merge

tests/e2e/specs/test.js
kenodressel: maybe we could merge https://github.com/aeternity/aepp-raendom/tree/feature/e2e-testing to this branch before we merge it to superhero

thepiwo: good for me

kenodressel:
I also have some issues (related to sass I think) where in the production version all `opacity: 100%` properties are compiled to `opacity: 1%`. Not sure whats causing this.

davidyuk:
> I also have some issues (related to sass I think) where in the production version all `opacity: 100%` properties are compiled to `opacity: 1%`. Not sure whats causing this.

By default, vue cli proposes to use `dart-sass`, it looks like a more modern alternative, but I have chosen `node-sass` because it was already used.

Though looks like you have fixed it by avoiding percentages. I have found a similar question on SO: https://stackoverflow.com/questions/59348502/vue-js-vue-cli-performing-npm-run-build-compiles-css-for-image-opacity-to-1

PR #141, opened 2020-03-16
display install wallet section only if not LoggedIn

### LINKED ISSUE #140: Install wallet still clickable overlaying trending

mradkov:
fixes #140


thepiwo:
should maybe be solved in the active css class though...

PR #148, opened 2020-03-17
Remove duplication of style definitions

### LINKED ISSUE #147: CSS import duplicates
base styles (base.scss+bootstrap) are included multiple times for each component resulting of duplication of rules and a css junk

A single point of style imports should be implemented or styles should be reorganized

venimus:
closes #147 

Proposal after research is to use sass-loader automatically include variables and mixins, which are required by components.

Including of `base.scss` everywhere results in adding the behemot bootstrap framework for each of the tens components

### DIFF
--- a/src/components/Dropdown.vue
+++ b/src/components/Dropdown.vue
@@ -22,7 +22,6 @@
 </script>
 
 <style lang="scss" scoped>
-  @import "../styles/base";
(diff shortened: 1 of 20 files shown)


venimus:
> * comments in profile seem broken now
seems style scopes are unexpectedly inconsistent. Would require a further fix

PR #167, opened 2020-03-19
Fix linter errors

CedrikNikita:


mradkov:
There are some conflicts in this PR. Can you resolve them?

thepiwo:
can you enforce linter pass in CI?

scripts/test.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+set -e
+
+npm run lint
```
thepiwo: should only be a lint check that fails CI if PR is not properly linted

.travis.yml
```diff
@@ -3,7 +3,9 @@ dist: xenial
 language: node_js
 node_js: '12'
 
-script: npm run build
+script: 
+  - scripts/test.sh
```
thepiwo: why not have it inside here?

venimus: Can these run in parallel? Imho it will save some time. Both can fail the build at any time.
Though usually the static checks should run first, but often single space fails to show you a real problem. So IMHO i would suggest parallel

https://docs.travis-ci.com/user/build-stages/#how-to-define-build-stages

thepiwo: I think it doesn't matter for our use-case

PR #176, opened 2020-03-19
Issue: #169 tip note mobile

stanislav-slavov:


src/components/tipRecords/TipComment.vue
```diff
@@ -15,7 +15,7 @@
             </span>
           </div>
         </div>
-          <div class="tip__note pr-2" :title="comment.text">
```
mradkov: Why don't we use `pr- / pl- / pt- / pb-` classes that comes from bootstrap? Or the padding here should be concrete number?

venimus: I think we don't even need bootstrap on this project.

stanislav-slavov: Because the styles is set with "!important" and it may differ on different resolutions.

stanislav-slavov: We started using it, because at the beginning we had no typography and the designs were provided as images. Now it is different.

mradkov: > I think we don't even need bootstrap on this project. It is almost not used anywhere

It would have been good to use its grid system for easier responsiveness.

stanislav-slavov: Should I use bootstrap padding here?

venimus: Yes, agree, but there is already implemented own logic for that and it became useless. At least the design is not bootstrap-grid based

stanislav-slavov: I guess we can remove this padding for now and decide on whether we are going to remove bootstrap from our implementation. This process could start in a separate issue where we remove its usage. Any thoughts?

PR #179, opened 2020-03-19
refactor ae symbol in AeAmount component

### LINKED ISSUE #172: refactor AE symbol to be in ae amount component

mradkov:
Fixes #172


stanislav-slavov:
Looks good. Few components were formatted. Shouldn't we do it for the whole site at a certain point, because it may cause a lot of conflicts.

thepiwo:
I agree with @stanislav-slavov here, formatting shouldn't be mixed with actual changes

src/components/layout/RightSection.vue
```diff
@@ -44,74 +43,74 @@
 </template>
 
 <script>
-import { mapGetters } from 'vuex';
-import { detect } from 'detect-browser';
-import FiatValue from '../FiatValue.vue';
-import AeAmount from '../AeAmount.vue';
-import Topic from '../tipRecords/Topic';
+import { mapGetters } from "vuex";
```
venimus: i think it should stay `'`

PR #181, opened 2020-03-19
Refactor layout

### LINKED ISSUE #144: Left panel inconsistencies
Left navigation panel sizes on higher resolutions (1366+) are inconsistent with design (dimmer bottom layer is the designed) :

1. Logo size
1. Font sizes/weight
1. Offset from the main panel
1. Padding between items
1. Offset of metrics

image

Current should stay for mid-res monitors (1024-1366)

### LINKED ISSUE #145: Main panel inconsistencies
Differences from the design (top is designed, bottom is implemented):
1. Body is not centered. All columns are little at right from the center
1. Offset from the right column is smaller

venimus:
closes #145
closes #144 

* Some general refactoring of the layout so it matches the design (center the 3 columns as whole instead around the middle column)
* Split the Left sections to 2 components
* Added the fonts which apparently were missing
* Reduce the number of breakopoints
* Fixed the font styles and paddings of the left section
* Some image adjustments


mradkov:
Problem with responsiveness: On resolution < 990px wide the side columns are not visible and the "center" feed is displayed on the right.

PR #183, opened 2020-03-20
Upload avatar

mradkov:


src/utils/backend.js
```diff
@@ -45,6 +45,21 @@ export default class Backend {
...
     return fetch(`${this.BACKEND_URL}/profile/image/${address}`);
   });
+  setProfileImage = async (address, data, image = true) => wrapTry(async () => {
+    if (image) {
+      return fetch(`${this.BACKEND_URL}/profile/image/` + address, {
+        method: 'post',
+        body: data,
+      });
+    }
+    else {
```
venimus: else is uneeded

src/utils/backend.js
```diff
@@ -45,6 +45,21 @@ export default class Backend {
...
+      return fetch(`${this.BACKEND_URL}/profile/image/` + address, {
```
venimus: address can go into the string template

src/views/UserProfileView.vue
```diff
@@ -322,7 +343,6 @@
         }
         .profile__image--edit{
           &>div{
-            position: absolute;
             top: 20%;
```
venimus: this is unneeded if the div is position:static

src/utils/backend.js
```diff
@@ -45,6 +45,21 @@ export default class Backend {
   getProfileImage = async (address) => wrapTry(async () => {
     return fetch(`${this.BACKEND_URL}/profile/image/${address}`);
   });
+  setProfileImage = async (address, data, image = true) => wrapTry(async () => {
+    if (image) {
+      return fetch(`${this.BACKEND_URL}/profile/image/` + address, {
```
venimus: address can go into the string template, also i would make the fetch params to be conditional, and call only one fetch

src/utils/backend.js
mradkov: It is needed as we send different `Content-Type` on the follow-up request. The avatar upload is 2 step process: uplaod image -> get challenge -> sign and respond.

src/views/UserProfileView.vue
mradkov: Change avatar text was not fully visible otherwise.

src/utils/backend.js
venimus: i mean, return will exit the function else will not be reached in this case.

mradkov: `image` flag is used to check what request are you sending to the backend .. if its `true` it gets in the first condition, otherwise it gets in the `else`.

venimus: I mean this is essentially the same
```
if (image) {
      return fetch(`${this.BACKEND_URL}/profile/image/` + address, {
        method: 'post',
        body: data,
      });
    }

      return fetch(`${this.BACKEND_URL}/profile/image/` + address, {
        method: 'post',
        body: JSON.stringify(data),
        headers: {'Content-Type': 'application/json'}
      });
```

venimus: image

mradkov: I've simplified it in https://github.com/aeternity/aepp-raendom/pull/183/commits/61eb92f9059ae6a6ca564de11ed49608a8ba7df5

src/views/UserProfileView.vue
```diff
@@ -154,9 +154,11 @@
         userCommentCount: 0,
         profile: {
           biography: '',
-          displayName: ''
+          displayName: '',
+          image: false,
         },
         avatar,
+        BACKEND_URL : "https://raendom-backend.z52da5wt.xyz",
```
thepiwo: why is the backend needed here?

src/views/UserProfileView.vue
```diff
@@ -227,16 +229,36 @@
...
+          }
+        })
+        .catch(console.error);
+      },
+      uploadImage(event) {
+        let data = new FormData();
+        data.append('name', 'image');
+        data.append('image', event.target.files[0]);
+
+        backendInstance.setProfileImage(this.account, data)
```
thepiwo: why not async await structure here?

venimus:
It works, but not really intuitive. Can it be made to preview the current avatar in editMode too (+showing the default image initially)?

src/views/UserProfileView.vue
mradkov: because the image is served from there.

mradkov:
> It works, but not really intuitive. Can it be made to preview the current avatar in editMode too (+showing the default image initially)? Or make the upload be triggered by click on the avatar (outside edit)

This can be done in a separate PR, I wanted to introduce the functionality here.

src/views/UserProfileView.vue
thepiwo: it should be fetched via backendInstance

mradkov: The default avatar will also be served via the backend, so I will refactor this.

src/utils/backend.js
```diff
@@ -26,7 +26,17 @@ export default class Backend {
...
+  getProfileImage = (address) => `${BACKEND_URL}/profile/image/${address}`;
+
+  setProfileImage = async (address, data, image = true) => wrapTry(async () => {
+    const request = {
+      method: 'post',
+      body: image ? data : JSON.stringify(data),
+    };
+    Object.assign(request, !image && { headers: { 'Content-Type': 'application/json' } });
+    console.log(request);
+    return fetch(`${BACKEND_URL}/profile/image/${address}`, request);
```
venimus:
```suggestion
    return fetch(getProfileImage(address), request);
```

src/utils/backend.js
```diff
@@ -26,7 +26,17 @@ export default class Backend {
     headers: { 'Content-Type': 'application/json' },
   }));
 
-  getProfileImage = async (address) => wrapTry(async () => fetch(`${BACKEND_URL}/profile/image/${address}`));
+  getProfileImage = (address) => `${BACKEND_URL}/profile/image/${address}`;
```
venimus: getProfileImageEndpoint?

src/views/UserProfileView.vue
```diff
@@ -302,6 +298,26 @@ export default {
         this.profile = profile;
       }).catch(console.error);
     },
+    getAvatar(address) {
+      return backendInstance.getProfileImage(address);
```
venimus: getProfileImageUrl

src/utils/backend.js
mradkov: Nope, this is the `post` method where we upload it.

venimus: Though it is a POST request, endpoints in json-apis are usually (should be) the same for get and post

PR #191, opened 2020-03-20
Better handling of missing images and upload avatar

venimus:
follow up to #190 

Still a hack, but I found this better working and cross-browser compatible, also easier to apply to other images. just use
```
<img :onerror="`this.style.opacity=0`">
```
it is inlined because dom is updated dynamically and to avoid overengeneering. 
```
<img :onerror="`this.error=null;this.src=${defaultImg}`">
```
 is an option but requires importing the asset. `this.error=null` is to avoid endless reload in case of defaultImg in turn fails to load (network issues).

I'm open to better ideas.

`defaultImg.svg` could be added as background of the container. in this case `img` itself should also have a css `background-color`, as pngs and svgs can be transparent and background will be visible too.

Applied to missing user avatar too

cleaned up the old hack


mradkov:

venimus:
Yes, will wait for Upload avatar first, because it will cause a conflict again

mradkov:
> Yes, will wait for Upload avatar first, because it will cause a conflict again

Its merged now, you can resolve.

src/views/UserProfileView.vue
```diff
@@ -35,12 +35,12 @@
                 :class="[showLoadingAvatar ? 'blurred' : '']"
               >
                 <a :href="openExplorer(address)" target="_blank" v-if="!editMode" :title="address">
-                  <img :src="getAvatar(address)" />
```
venimus: self-closing tags are not valid html5

mradkov: isn't this is enforced by the linter?

venimus: i don't think it even checks the html

PR #198, opened 2020-03-21
Added retip icon in deeplinks implementation

stanislav-slavov:


src/components/tipRecords/FormatDate.vue
```diff
@@ -0,0 +1,18 @@
+<template>
+  <span>{{formatDate}}</span>
```
venimus: Can you change it to something like this
```
  <time datetime="2008-02-14 20:00" title="2008-02-14 20:00">{{ formatDate }}</time>
```

venimus: Title is very helpful to get the exact time, because now it will will be hidden after 1 day

src/components/tipRecords/FormatDate.vue
```diff
@@ -0,0 +1,21 @@
...
+  props: ['dateTimestamp'],
+  computed: {
+    formatDate() {
+      const today = new Date();
+      const tipDate = new Date(this.dateTimestamp);
+      const isToday = (today.toDateString() === tipDate.toDateString());
+      return isToday ? tipDate.toLocaleTimeString('en-US', { hourCycle: 'h24' }) : tipDate.toLocaleDateString('en-US');
+    },
+    wholeDateAndTime() {
+      return new Date(this.dateTimestamp).toLocaleString('en-US', { hourCycle: 'h24' });
```
venimus: `datetime="YYYY-MM-DDThh:mm:ss"` should be machine readable which is Date().toISOString() while title should be human readable localeString of the full date

PR #202, opened 2020-03-23
Issue #199 Removed mission page

stanislav-slavov:


src/views/Privacy.vue
```diff
@@ -2,7 +2,7 @@
   <div>
     <right-section></right-section>
     <left-section></left-section>
-    <div class="container wrapper">
+    <div class="container wrapper mt-2">
```
venimus: IMHO we should later remove that container.wrapper stuff and replace it with something proper also remove boostrap at all

stanislav-slavov: I agree we can do it at a later stage

PR #204, opened 2020-03-23
Load data from cache

thepiwo:
in combination with https://github.com/aeternity/tipping-community-backend/pull/64


mradkov:
Since this is non-blocking I suggest we merge it, and have everything prepared for when the backend is also updated? cc @thepiwo

thepiwo:
Let's wait for review from keno

Issue #171, opened 2020-03-19
Tip section is shown without having active account in wallet


mradkov:
The tip section is shown when user has installed the superhero wallet but has not yet created a keypair.

I think this should not be the case - meaning it should be only shown when the wallet is ready for tipping.

Currently this results to:

<img width="1431" alt="Screenshot 2020-03-19 at 13 00 43" src="https://user-images.githubusercontent.com/13139371/77060860-efcd7900-69e1-11ea-8e08-2e1c6c87e146.png">

etharner:
@mradkov Shouldn't this issue be closed now?

PR #209, opened 2020-03-24
Users avatars, minor css adjustments, fixes

### LINKED ISSUE #184: Display user avatar
Display user avatar everywhere where needed.

We should think of more performant way of displaying avatars, as currently we need to make 1 request to check if the user has avatar and then fetch the image with second request.

Maybe we can add the default avatar in the backend and have it displayed for every user that has no avatar (currently we keep it as local asset) - this way we will only make a single request for getting avatar image per user.

UPDATE: The backend now serves a fallback avatar image for user profile without custom avatar, so we are free to directly get and show the avatars from there.

mradkov:
- show user avatar dynamically (closes #184)
    - have getProfileImageUrl as `static` method as well
- adjust css for displaying avatar properly
- minor fixes:
    - onboarding wizzard tutorial link
    - onbosarding wizzard to be only visible if no wallet
    - trending leads to the main tiplistview

### DIFF
--- a/src/components/layout/SendTip.vue
+++ b/src/components/layout/SendTip.vue
@@ -69,6 +69,7 @@ import util from '../../utils/util';
 import aeternity from '../../utils/aeternity';
 import { EventBus } from '../../utils/eventBus';
 import avatar from '../../assets/userAvatar.svg';
+import Backend from '../../utils/backend';
 
 export default {
   name: 'SendTip',
@@ -76,7 +77,7 @@ export default {
     FiatValue,
   },
   computed: {
-    ...mapGetters(['balance', 'loading']),
+    ...mapGetters(['balance', 'loading', 'account', 'isLoggedIn']),
     isSendTipDataValid() {
       const urlRegex = /(https?:\/\/)?([\w-])+\.{1}([a-zA-Z]{2,63})([/\w-]*)*\/?\??([^#\n\r]*)?#?([^\n\r]*)/g;
       // TODO: better validation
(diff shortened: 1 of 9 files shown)


src/components/onboarding/steps/Step2.vue
```diff
@@ -12,8 +12,8 @@
     <p>
       Fund transfers are instant, with near-zero processing fees.
       Censorship-proof: there are no third-party intermediaries to ban creators from sending and
-      receiving payments, and records are immortalized on æternity blockchain. Learn more by our
-      Tutorial & FAQ
+      receiving payments, and records are immortalized on æternity blockchain. <a href="https://youdonotneedacapetobeahero.com/#/create-profile" target="_blank">Learn more by our
```
venimus: router?

mradkov: Its intentionally by designers request like this - so you don't leave the wizzard.

venimus: I guess we can use something like this to avoid bugs after domain migration
```
<router-link to="/create-profile" v-slot="{ href }"><a :href="href" "target=_blank">Tutorial & FAQ</a></router-link>
```

src/components/layout/SendTip.vue
```diff
@@ -106,6 +107,17 @@ export default {
     clearTipForm() {
       this.sendTipForm = { amount: null, url: '', title: '' };
     },
+    getAvatar(address) {
+      return Backend.getProfileImageUrl(address);
+    },
+  },
+  async created() {
+    const loadUserAvatar = setInterval(() => {
```
venimus: what is the issue here

mradkov: Its related to #171.

PR #224, opened 2020-03-24
display send tip section only when user has active account

mradkov:
Fixes #175 in a hacky way.


mradkov:
> Can you resolve the problem with the tests?

I think the problem was 2 of the tests were timing out. Locally they work fine. I've extended the timeout for them. Lets see if it works in travis.

Issue #212, opened 2020-03-24
Tip and retip buttons adjustments


mradkov:
Tip and retip currently works only by clicking on the icon (shown in pink). I suggest we extend the clicking area (shown in red) for easier accessibility.

<img width="546" alt="Screenshot 2020-03-24 at 9 24 09" src="https://user-images.githubusercontent.com/13139371/77399102-69ce7b00-6db1-11ea-9e3e-086f42d4b452.png">

PR #226, opened 2020-03-25
Issue #212 Adjusted tip/retip buttons texts to be clickable 

stanislav-slavov:


venimus:
I think the PR title is misleading

Issue #219, opened 2020-03-24
Hover states of tip blocks


venimus:
Desktop only,  there are basically 2 levels of visual feedback to the user:
1. The mouse is within the tip block (even the blank space) - make the wrapper background a little lighter (increase color luminocity to 18%) 
2. The mouse is hovering an element within the tip block - additional effect specific to it, i.e. `1.` still applies
    * Put underline on `#tags` on hover
    * Make the inner site preview section background a little lighter when mouse is hovered on it + lighter text + add underline to the domain. 
    * Add hover state with lighter color on the user address/name

Mobile:
* Make only tags and domain underlined permanently (desktop hover stuff does not apply)

Synced with designers

venimus:
closed via #232

PR #232, opened 2020-03-25
Align Tip Records view with design and polishing

venimus:
+ polish a little the img fallback
+ native browser lazy loading of images
basically add `loading="lazy"` to the tip images which most modern browsers support (though it is still "[experimental](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img)")
+ added hover states of tip blocks #219 
+ Other minor visual issues
+ clean and format some of the css (TipRecord.vue)


src/styles/base.scss
```diff
@@ -103,10 +103,18 @@ a:hover{
   }
 }
 
+img.fail{
```
CedrikNikita:
```suggestion
img.fail {
```
Will be awesome if you will choose rather you have this space before the brace in styles or not. Because most of this document have it. You have different style of spacing in your own PR.

venimus: Yes, accidentally. It should have a space, I also recommend adding a linter for scss
I avoid formatting the old files, so i do it manually, so sometimes miss it.

mradkov:
Can you also resolve the conflicts?

Issue #248, opened 2020-03-27
If there is no link preview, the link is not visible at all


kenodressel:
For example:
![image](https://user-images.githubusercontent.com/765921/77749886-1b281780-7023-11ea-9cbb-bd7c621bf3b7.png)

The link should be displayed, even if there is no preview.

mradkov:
This was fixed here: https://github.com/aeternity/aepp-raendom/pull/203

apparently someone did overwrite it..

venimus:
I just had formatted it https://github.com/aeternity/aepp-raendom/pull/232/files#diff-0c9c134dbd57fccd61d57dc7ad9fac6fR120-R126

Imho it looks like this case wasn't covered

venimus:
I think the problem is that there was no `http://` in the url when submitted by the user. The parser couldn't access it, then failed to generate any preview out of it. I think the backend should prepend the missing prefixes before trying to access the site, also frontend should prefix the href with //

mradkov:
There always should be `http/https` prefix on the URL sent for tipping.

venimus:
Yes, it appears the validation failed for some reason probably due to a bug and there is now invalid data in the db.

PR #251, opened 2020-03-27
Issue #248 Show link when there is no preview

stanislav-slavov:


src/components/tipRecords/TipRecord.vue
```diff
@@ -282,6 +285,16 @@ export default {
     }
   }
 
+  .tip__url {
+    margin-left: 1rem;
+    margin-right: 1rem;
+
+    a {
```
venimus: can you add :hover - underline for this link, please to maintain consistency with the other preview

src/components/tipRecords/TipRecord.vue
```diff
@@ -456,6 +469,10 @@ export default {
       }
     }
 
+    .tip__url{
```
venimus:
```suggestion
    .tip__url {
```

venimus:
This is acceptable but can not fix the issue in general, as the bug is complex

PR #255, opened 2020-03-27
Maintenance page and display errors with connection and backend on it

vmangelovv:


.gitignore
```diff
@@ -1,28 +0,0 @@
-.DS_Store
```
venimus: I meant to ignore .history, not to remove the file

src/views/TipCommentsView.vue
```diff
@@ -48,11 +48,13 @@
 
 <script>
 import { mapGetters } from 'vuex';
+// eslint-disable-next-line import/no-cycle
```
venimus: Can we avoid these somehow @mradkov ?

vmangelovv: I would ask about that too..

src/utils/aeternity.js
```diff
@@ -95,8 +96,13 @@ aeternity.initClient = async () => {
     try {
       aeternity.client = await aeternity.initStaticClient();
       result = await aeternity.initProvider();
-    } catch (e) {
-      console.error(e);
+    } catch (err) {
+      router.push({
```
venimus: i think this redirect could become annoying for users. Imagine typing something and you get an error, you will lose what you were doing. I think might be better to show some popup or similar, so user could keep their position (or whatever they do).

vmangelovv: Ok, but what is the point of new static page then?

venimus: i don't actually see a point, but could be a temporary solution for the end-users for now

src/views/TipCommentsView.vue
mradkov: They are not present in https://github.com/aeternity/aepp-raendom/pull/255/commits/81204d1b339216cbf0832f61fe3f0a009913fcf7 as I can see?

vmangelovv: -> 8f26461

src/App.vue
```diff
@@ -98,6 +98,14 @@ export default {
     },
   },
   async created() {
+    EventBus.$on('redirectMaintenance', (err) => {
```
venimus: Event names should represent what happened (not how it will be handled), so I guess this should be something like 'backendError'

venimus:
I think this is good enough, just name the event

mradkov:
> btw, didn't test it

I'll do it now.

mradkov:
I'm not sure how helpful it will be for the user to dump the error in the page?

Shouldn't we just have 2 sentences explaining there is an error (connectivity/other) instead?

Screenshot 2020-03-30 at 17 25 48

src/views/Maintenance.vue
```diff
@@ -0,0 +1,89 @@
+<template>
+  <div>
+    <div class="maintenance container wrapper mt-2">
```
venimus, line 3: could you please do not use mt-2, pt-2, text-center and other bootstrap classes (Bootstrap is useless, we would like to clean up and remove it). Instead add the properties to the specific class (add one if missing). In this case add margins to `.maintainance`. You can leave `.container`+`.wrapper` if needed.

src/views/Maintenance.vue
```diff
@@ -0,0 +1,89 @@
+<template>
+  <div>
```
venimus, line 2: This div is redundant, please remove it.

src/views/Maintenance.vue
```diff
@@ -0,0 +1,89 @@
...
+    <div class="maintenance container wrapper mt-2">
+      <div class="p-5 content text-center">
+        <div v-if="view === 'recharge'">
+          <div class="recharge">
+            <h1>We're recharging our powers.</h1>
+            <div class="gif" />
```
venimus, line 8: why not simply use img? using background images instead of image comes with a lot of trade-offs, best practice is to avoid it. To center the image inside the constraint/fixed size, add a class to it and apply a property i.e. `object-fit: cover`. Also please use more relevant naming - i.e. maintenance--image (though naming convention is not yet agreed). Anyway .gif, .gif2 is quite inappropriate, image format might change, and even now .gif2 contains a svg.

src/views/Maintenance.vue
```diff
@@ -0,0 +1,89 @@
...
+      view: 'recharge',
+    };
+  },
+};
+</script>
+
+
+<style lang="scss" scoped>
+.recharge {
+  h1 {
```
venimus, line 39: please, put the repeating styles (h1, h2) in a parent selector (.maintenance)

src/views/Maintenance.vue
```diff
@@ -0,0 +1,89 @@
...
+    line-height: 80px;
+    margin: 50px 0;
+  }
+  .gif {
+    background: url('../assets/recharge.gif') center center;
+    display: block;
+    height: 190px;
+    width: 208px;
+    margin: 0 auto;
+    position: relative;
```
venimus, line 53: this seems redundant

Issue #230, opened 2020-03-25
Style Tutorial and FAQ pages as per latest designs


stanislav-slavov:

venimus:
closed via #246

PR #256, opened 2020-03-27
Issue #230 Styled Tutorial page

stanislav-slavov:


venimus:
could you please add ref #230 or closes #230 in description, so github can handle relations

PR #259, opened 2020-03-30
Fix tests

thepiwo:


mradkov:
> Post transaction is failing to the https://testnet.aeternity.io
> While calling postTransaction (body), POST to https://testnet.aeternity.io/v2/transactions failed with 400: Invalid tx
> at Object.g [as sendTransaction]
> Also i suggest to use timeout instead of wait according to the cypress best practice, in case out latency will decrease, also loading time may change due to pagination feature.

Can you resolve those?

thepiwo:
@CedrikNikita is there any progress here?

mradkov:
@CedrikNikita Can you resolve the conflicts and test and fix the failing tests?

PR #263, opened 2020-03-31
Created Mobile Navigation

stanislav-slavov:
ref #258


src/components/layout/MobileNavigation.vue
```diff
@@ -0,0 +1,175 @@
+<template>
+  <div
+    class="mobile-navigation clearfix"
+    :class="[
```
venimus: i would prefer using object
```
:class="{
   open: open
   sticky: $route.name !== 'home'
}"

File: src/components/layout/MobileNavigation.vue [hunk 2]
```diff
@@ -0,0 +1,175 @@
...
+    >
+      &#x2715;
+    </div>
+    <navigation/>
+    <div class="mobile-actions">
+      <img
+        class="tip"
+        :src="emptyHeart"
+        @click="toggleTipModal(true)"
+        v-if="!open && $route.name === 'home'"
```
venimus: you can move the v-if=!open part in the parent element, seems more logical to me

stanislav-slavov: The trigger-navigation button has a different condition. There are 2 buttons that disappear if   we are not on the home page. The parent would hide all and is used for styling.

src/components/layout/MobileNavigation.vue
```diff
@@ -0,0 +1,175 @@
...
+      >
+      <img
+        class="trigger-navigation"
+        :src="iconMobileMenu"
+        @click="openNavigation(true)"
+        v-if="!open"
+      >
+    </div>
+    <div class="footer text-center" v-if="open">
```
venimus: please don't use text-center (or similar), there is a custom class already (.footer) just add centering there.

src/components/layout/MobileNavigation.vue
```diff
@@ -0,0 +1,175 @@
...
+      SuperHero is Open Source
+      <router-link class="ae" to="/terms">
```
venimus: can you add some specific class name to this, .ae is used elsewhere for styling the AE currency

src/components/layout/MobileNavigation.vue
```diff
@@ -0,0 +1,175 @@
...
+      </router-link>
+    </div>
+    <mobile-send-tip-modal></mobile-send-tip-modal>
+  </div>
+</template>
+
+<script>
+import { mapGetters, mapActions } from 'vuex';
+import Navigation from './Navigation.vue';
+import iconMobileMenu from '../../assets/iconMobileMenu.svg';
```
venimus: i think this import is useful only for dynamic assets. You can simply put that in src="" (without `:`).

src/components/layout/MobileNavigation.vue
venimus: i mean only !open but leave the route === home per item where applies

stanislav-slavov: Yes, you are right

src/views/TipsList.vue
```diff
@@ -73,10 +77,12 @@ import TipRecord from '../components/tipRecords/TipRecord.vue';
 import SendTip from '../components/layout/SendTip.vue';
 import LeftSection from '../components/layout/LeftSection.vue';
 import RightSection from '../components/layout/RightSection.vue';
+import MobileNavigation from '../components/layout/MobileNavigation.vue';
 import { EventBus } from '../utils/eventBus';
 import Loading from '../components/Loading.vue';
 import Onboarding from '../components/onboarding/Wizard.vue';
 import { MIDDLEWARE_URL } from '../config/constants';
+import iconEraser from '../assets/iconEraser.svg';
```
venimus: you missed one

stanislav-slavov: Thanks. Also I will restore the back arrow for other resolutions than mobile in the next commit

src/store/index.js
```diff
@@ -30,6 +30,8 @@ export default new Vuex.Store({
     chainNames: [],
     wizardCurrentStep: 0,
     wizardIsCollapsed: false,
+    hideMobileNavigation: false,
```
venimus: do we need to persist the state of the navigation?

stanislav-slavov: It is implemented like so - if we have to collapse the mobile menu from a different pages and components. In this case we have the search.

src/views/CreateProfile.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <div>
+    <mobile-navigation></mobile-navigation>
```
venimus: this needs to evolve in a Layout component some time in the near future

mradkov:
Can you resolve the conflicts @stanislav-slavov ?

stanislav-slavov:
> Can you resolve the conflicts @stanislav-slavov ?

Done.

mradkov:
Still some conflicts.

stanislav-slavov:
> Still some conflicts.

Rebased

PR #269, opened 2020-04-01
Updated FAQ and Tutorial pages, restructured the expandable block component

stanislav-slavov:


src/components/ExpandableBlock.vue
```diff
@@ -8,27 +8,23 @@
...
+    <div class="title">
+      <slot name="title" />
+    </div>
     <div
       v-if="expanded"
       class="body"
-      v-html="body"
-    />
+    >
+      <slot name="body" />
```
venimus, line 18: trick: if you do not specify a `name` but simply use `<slot>` then it will render the content block that is not wrapped with `<template slot=..>`

```
<expandable-block>
  <template slot="title">
    HOW TO USE SUPERHERO
  </template>
  <p><b>Download and install the Superhero extension for your browser.</b></p><p /><p>
...
```
instead of

```
  <template slot="title">
    HOW TO USE SUPERHERO
  </template>
  <template slot="body">
    <p><b>Download and install the Superhero extension for your browser.</b></p><p /><p>
...
  </template>
```
it will save a lot of `<template>` tags and reduce clutter

venimus: Above could be further cleaned up if you replace the title slot with a component property

src/views/CreateProfile.vue
```diff
@@ -10,359 +10,290 @@
...
+              nor connect any password with any account.<br>
+              The task of keeping the keys to your wallet is yours and yours alone, Superhero.
+            </p>
+            <p>
+              Here’s how to secure and backup your wallet:<br>
+              Once you’re in your dashboard, you’ll get a series
+              of words called a seed phrase.<br>
+              The seed phrase is somewhat like a password –
+              it’s a unique set of words (in order) which connects you to your wallet.<br>
+              <ul>
```
venimus, line 93: semantically `<ul/ol>'s are not supposed to be inside `<p>` though browsers do not usually complain

PR #271, opened 2020-04-02
travis: Build stage version

davidyuk:


.travis.yml
```diff
@@ -10,14 +10,25 @@ script:
...
+    skip_cleanup: true
+    github_token: $GITHUB_TOKEN
+    keep_history: true
+    local_dir: dist
+    repo: $DEPLOY_REPO
+    target_branch: master
+    name: $DEPLOY_NAME
+    email: $DEPLOY_EMAIL
+    on:
+      branch: superhero-production
```
davidyuk, line 23: The only problem is branch naming. Should we merge `feature/layout-superhero` to `master`? This line proposes to use `superhero-production` as the temporary master branch.

mradkov: We can keep it `superhero-production` for now as you suggested and merge in `master` after we have agreement.

PR #278, opened 2020-04-03
travis: Add missed all_branches: true

davidyuk:


davidyuk:
In this case, Travis wrote:
> Skipping a deployment with the pages provider because a custom condition was not met

Should work on ` feature/layout-superhero`/`superhero-production`
@mradkov

.travis.yml
```diff
@@ -20,4 +20,5 @@ deploy:
   name: $DEPLOY_NAME
   email: $DEPLOY_EMAIL
   on:
+    all_branches: true
     condition: $DEPLOY_REPO
```
davidyuk, line 24: the condition is that $DEPLOY_REPO defined

Issue #282, opened 2020-04-03
wizzard create profile broken if no extension installed


thepiwo:
![image](https://user-images.githubusercontent.com/6539235/78361126-27c6e580-75b8-11ea-9447-9280c7c24a78.png)

PR #283, opened 2020-04-03
Changed texts, fixed issue with profile link on the Onboarding Wizard

stanislav-slavov:
ref #282

### DIFF
--- a/src/components/onboarding/Wizard.vue
+++ b/src/components/onboarding/Wizard.vue
@@ -131,10 +131,10 @@ export default {
       this.setWizardCurrentStep(index);
     },
     finalStep() {
-      // set step after the last to hide permanently (or until a new 'step' is added eventually):
       this.setWizardIsCollapsed(true);
 
       if (this.account && this.account.length) {
+        // set step after the last to hide permanently (or until a new 'step' is added eventually):
         this.setWizardCurrentStep(steps.length);
         this.$router.push({
           name: 'user-profile',
--- a/src/components/onboarding/Wizard.vue
+++ b/src/components/onboarding/Wizard.vue
@@ -134,10 +134,8 @@ export default {
       // set step after the last to hide permanently (or until a new 'step' is added eventually):
       this.setWizardIsCollapsed(true);
 
-      this.setWizardCurrentStep(steps.length);
-
-
-      if (this.account) {
+      if (this.account && this.account.length) {
+        this.setWizardCurrentStep(steps.length);
         this.$router.push({
           name: 'user-profile',
           params: {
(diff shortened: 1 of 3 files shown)


src/components/onboarding/steps/Step3.vue
```diff
@@ -19,6 +19,53 @@
       People can directly support your content, cause project<br>
       (or you for simply being you) with no middleman.
     </p>
+
+    <p class="highlited">
+      To start, you'll need a wallet. It's free. It's easy.<br>
+      <span class="important">
+        1. Click a browser extensions or mobile app
```
venimus, line 26: This `1.` when there is no 2 is very confusing

src/components/onboarding/Wizard.vue
```diff
@@ -134,10 +134,8 @@ export default {
       // set step after the last to hide permanently (or until a new 'step' is added eventually):
       this.setWizardIsCollapsed(true);
 
-      this.setWizardCurrentStep(steps.length);
-
-
-      if (this.account) {
+      if (this.account && this.account.length) {
+        this.setWizardCurrentStep(steps.length);
```
venimus, line 138: Why is that? this will break the wizard from going forward. If it is intentional, changing steps from the top navigation will still bypass this logic

src/components/onboarding/Wizard.vue
```diff
@@ -134,10 +134,8 @@ export default {
       // set step after the last to hide permanently (or until a new 'step' is added eventually):
```
venimus, line 134: this comment is misplaced it should be for `this.setWizardCurrentStep(steps.length);` so could you please move it into the `if` too

src/components/onboarding/steps/Step3.vue
mradkov: It was the wording provided from comms i guess.

mradkov:
Looks good though

src/components/onboarding/steps/Step3.vue
venimus: Yes, I think it is just a typo from the comms

stanislav-slavov: Yes. This is how the content.

src/components/onboarding/Wizard.vue
stanislav-slavov: Sure

src/components/onboarding/steps/Step3.vue
mradkov, line 26:
```suggestion
        Click a browser extensions or mobile app
```

PR #284, opened 2020-04-03
Added missing texts from FAQ

stanislav-slavov:


src/views/FAQ.vue
```diff
@@ -133,6 +136,18 @@
           Currently, users can only add short descriptions for their profile.
           But don’t despair—profile photos and avatars are coming soon!
         </expandable-block>
+        <expandable-block :title="'Do I have to add my wallet address to every Tweet?'">
+          Yes. At the moment, the wallet is not able to fetch the wallet address from
+          a Twitter profile through a Tweet.  So you would have to add your wallet
+          address to every Tweet. Hopefully, this will be improved later.
+        </expandable-block>
+        <expandable-block
+          :title="`If the owner of the URL I tipped doesn’t have an
```
venimus, line 145: I think this should also have `'` and be like that 
```
:title="`'If the 
```

stanislav-slavov: Then it will be displayed. `` makes it a string and available for mulitiple lines

Issue #293, opened 2020-04-06
Profile address links to mainnet.aeternity.io but should be mainnet.aeternal.io


kenodressel:
mainnet.aeternity.io
--> node / middleware

mainnet.aeternal.io
--> explorer

venimus:
![image](https://user-images.githubusercontent.com/632282/78558882-1120c880-781c-11ea-9ad0-82f1d4a12000.png)

PR #295, opened 2020-04-06
Changed explorer link

stanislav-slavov:
ref #293 
OP issue: #650


src/views/UserProfileView.vue
```diff
@@ -305,7 +304,7 @@ export default {
   },
   data() {
     return {
-      explorerUrl: `${MIDDLEWARE_URL}account/transactions/`,
```
thepiwo, line 308: should probably still be within constants

PR #299, opened 2020-04-06
Stats block is visible on wider screens

stanislav-slavov:
ref #298
OP task: #603

### DIFF
--- a/src/components/layout/Overview.vue
+++ b/src/components/layout/Overview.vue
@@ -165,25 +165,14 @@ export default {
     cursor: pointer;
   }
 
-  @media only screen and (min-width: 1440px) {
+  @media only screen and (max-width: 1280px) {
     .overview__value {
-      font-size: 1.6rem;
-      line-height: 1.6rem;
-      padding-bottom: 0.3rem;
-    }
-
-    .overview__label {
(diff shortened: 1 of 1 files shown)


venimus:
Btw https://github.com/aeternity/superhero-ui/pull/299/files#diff-64fb8a23b2e52577d4721f1f3eb493f6R156 (`.overview__label`) seems this was a wrong value initially. It should be `$tip_note_color`. Can you change it above (can't comment it) in the scope of this issue.

Else I think designers have to check that afterwards, because I suspect font sizes (spacing) should have some who-knows proportions (VALUE/label).

stanislav-slavov:
> Btw https://github.com/aeternity/superhero-ui/pull/299/files#diff-64fb8a23b2e52577d4721f1f3eb493f6R156 (`.overview__label`) seems this was a wrong value initially. It should be `$tip_note_color`. Can you change it above (can't comment it) in the scope of this issue.
> 
> image
> 
> 1. Designed   2. Implemented
> 
> Else I think designers have to check that afterwards, because I suspect font sizes (spacing) should have some who-knows proportions (VALUE/label).

Added $tip_note_color

PR #300, opened 2020-04-07
Add css linter

CedrikNikita:


src/views/TipsList.vue
```diff
@@ -138,144 +138,149 @@ export default {
...
   }
-  .tips__container{
-    width: 100%;
-    background-color: $background_color;
-    padding-top: 0.1rem;
-    &:empty{
-      visibility: hidden;
-    }
+
+  z-index: 100;
```
venimus, line 178: this looks weird, it should actually be grouped with the other properties. Is it possible to force that

CedrikNikita: Old:
https://github.com/aeternity/superhero-ui/blob/master/src/views/TipsList.vue#L177
New:
https://github.com/aeternity/superhero-ui/blob/bbfb603c5147622c4920344b03b9f82d67bc6b5e/src/views/TipsList.vue#L178
I don't undestand your point.

CedrikNikita: Actually i think we can add an issue with such rule, cause i haven't found it yet and i don't wanna get stuck with this PR due to a lot of conflicts it leads to, if othere PRs will be merged.
Also it should be discussed with everyone along with [this rule](https://github.com/aeternity/superhero-ui/pull/300/files#diff-3d9639f99b2cd6bda90b892bea090befR5). I want to add this rule as an issue and discuss in the comments there.

venimus: It's a good start tough

mradkov:
can you resolve the conflicts @CedrikNikita

PR #301, opened 2020-04-07
Back button spacing

stanislav-slavov:
OP bug: #655

### DIFF
--- a/src/views/UserProfileView.vue
+++ b/src/views/UserProfileView.vue
@@ -611,7 +611,7 @@ export default {
     background-color: $actions_ribbon_background_color;
     position: sticky;
     z-index: 21;
-    top: 3rem;
+    top: 3.1rem;
     a {
       font-weight: 600;
       color: $light_font_color;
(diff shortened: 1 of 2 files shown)


src/views/UserProfileView.vue
```diff
@@ -690,7 +690,7 @@ and (max-device-width: 480px)
 and (-webkit-min-device-pixel-ratio: 2) {
   .profile__page {
     .profile__actions {
-      top: 3.3rem;
+      top: 3.1rem;
```
venimus, line 693: This should be 3rem, because there will appear a gap

image

If you want to have some "spacing" I suggest adding a top-border with a background color

venimus:
Can you also remove `padding-top: 0.75rem;` from `.profile__page` as by design middle page should start from the top (no top padding), in addition when scrolled down it have some annoying "collapse" effect of the 'black section' (this is because `position: sticky` can not be used with "padding/margin")

PR #303, opened 2020-04-07
Display warning when user opens superhero from a browser which we don't support

vmangelovv:


src/App.vue
```diff
@@ -2,32 +2,47 @@
...
     return {
       page: 1,
+      browser: detect(),
+      alert: false,
     };
   },
   computed: {
     ...mapGetters(['settings', 'tipSortBy']),
+    supportedBrowsers() {
+      return ['chrome', 'firefox'];
```
venimus, line 37: i think all chromium browsers should be supported (i.e. Opera, Brave, even New Edge),

src/App.vue
```diff
@@ -2,32 +2,47 @@
...
+    },
   },
   async created() {
+    if (!IS_MOBILE_DEVICE && this.browser) {
```
venimus, line 41: I would have put all that in 'computed' it is just a 'return' value
like:
```
isSupportedBrowser(){
const browser = detect();
return IS_MOBILE_DEVICE || (browser && supportedBrowsers.includes(browser.name));
}
```
also `const supportedBrowsers = [...] ` should be put before(outside) `export default {` (tough best would be to have that imported from some const file, but not sure we use that elsewhere)

src/App.vue
mradkov: You are generally correct but we haven't actually tested the extension on Opera and the New Edge, so its better to warn people than to mislead them at this point.

src/App.vue
```diff
@@ -26,6 +31,10 @@ export default {
   },
   computed: {
     ...mapGetters(['settings', 'tipSortBy']),
+    isSupportedBrowser() {
+      const browser = detect();
+      return !IS_MOBILE_DEVICE && (browser && supportedBrowsers.includes(browser.name));
```
venimus, line 36: I think mobile should be "supported". if "yes" should revert the check and use ||

venimus:
I think we should not update `package-lock.json`, without changing `package.json`, or if having any issues with packages. Else, please revert if it is unrelated change.

src/App.vue
```diff
@@ -2,20 +2,25 @@
   <div id="app">
     <div
       ref="wrapper"
-      class="min-h-screen wrapper"
-    >
```
venimus, line 6: though not an error, but linter complains - `>` should be on a new line if attributes are multilined
```
> vue-cli-service lint --no-fix
warning: Expected 1 line break before closing bracket, but no line breaks found (vue/html-closing-bracket-newline) at src/App.vue:5:35:
  3 |     <div
  4 |       ref="wrapper"
> 5 |       class="min-h-screen wrapper">
    |                                   ^
  6 |       <div
  7 |         v-if="!isSupportedBrowser"
  8 |         style="text-align: center;line-height: 5rem;">
warning: Expected 1 line break before closing bracket, but no line breaks found (vue/html-closing-bracket-newline) at src/App.vue:8:54:
   6 |       <div
   7 |         v-if="!isSupportedBrowser"
>  8 |         style="text-align: center;line-height: 5rem;">
     |                                                      ^
   9 |         Unsupported browser.
  10 |       </div>
  11 |       <router-view />

File: src/App.vue [hunk 5]
```diff
@@ -2,20 +2,25 @@
...
+      class="min-h-screen wrapper">
+      <div
+        v-if="!isSupportedBrowser"
+        style="text-align: center;line-height: 5rem;">
```
venimus, line 8: please add this in `<style>`

src/App.vue
venimus: I actually use mainly Opera (among the others), it works perfectly. Tough on the chrome-Edge the site works, you just cant install chrome extensions, so not sure what "unsupported should mean".

venimus: btw in create-profile help it is mentioned Brave and Opera
image
So I guess, showing "unsupported browser" might be awkward

src/App.vue
vmangelovv: on mobile we use deep links so i am not sure if it should be supported.
@mradkov ?

mradkov: On mobile we do not need to show this.

vmangelovv:
Accidentally closed.
Ref. https://github.com/aeternity/superhero-ui/pull/313

PR #307, opened 2020-04-07
Changed favicon

stanislav-slavov:
ref #306
OP task: #663


venimus:
I think we should keep 2 separate images. May be just duplicate favicon.svg as iconExtension.svg.

PR #310, opened 2020-04-07
Update FAQ

mradkov:
- Update FAQ section to reflect the twitter parser
- Updated texts in FAQ
- updated texts in tutorial

### DIFF
--- a/src/views/FAQ.vue
+++ b/src/views/FAQ.vue
@@ -16,7 +16,7 @@
         </p>
         <expandable-block :title="'What is Superhero?'">
           Superhero is a P2P (Peer-to-Peer) platform for social sharing.
-          <br>The mission of Superhero is to empowering people to support people,
+          <br>The mission of Superhero is to empower people to support people,
           without intermediaries.
           <br>
           <router-link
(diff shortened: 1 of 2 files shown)


src/views/FAQ.vue
```diff
@@ -9,17 +9,16 @@
...
-          to<br> receive more information about using Superhero platform and wallet.
+          </router-link>to
+          <br>receive more information about using Superhero platform and wallet.
         </p>
         <expandable-block :title="'What is Superhero?'">
-          Superhero is a P2P (Peer-to-Peer) platform for social sharing.<br>
-          The mission of Superhero is to empowering people to support people,
-          without intermediaries.<br>
+          Superhero is a P2P (Peer-to-Peer) platform for social sharing.
+          <br>The mission of Superhero is to empowering people to support people,
```
venimus, line 19: imho, this should be
```suggestion
          <br>The mission of Superhero is to empower people to support people,
```

mradkov: I agree. These were the texts provided by the comms. Wasnt expecting to have an error there but 🤷‍♀ ..

PR #314, opened 2020-04-08
Add default account identicon

petbaik:
Not sure if covered all usages of avatar.


src/views/UserProfileView.vue
```diff
@@ -443,7 +441,7 @@ export default {
 </script>
 
 
-<style lang="scss" scoped>
+<style lang="scss">
```
mradkov, line 444: Looks good to me. I'm not really sure if we should remove `scoped` here or add the style to the Avatar component.

cc @venimus what do you think?

src/components/Avatar.vue
```diff
@@ -0,0 +1,49 @@
+<!-- eslint-disable-next-line vue/no-v-html -->
+<template>
+  <img
+    v-if="!error"
+    :src="avatar"
+    @error="error = true"
```
venimus, line 6: I think this will cause to trigger jdenticon regenerating on every placeholder, tough might not be so much. However it will still make a doomed request (that will end of 404) from the browser. Guess it will be goot to have a fallback (like unexpected server issue), but I believe it is not good to have a main logic that is driven by errors.

src/components/Avatar.vue
```diff
@@ -0,0 +1,49 @@
...
+    },
+  },
+  data: () => ({
+    defaultAvatar,
+    error: false,
+  }),
+  computed: {
+    avatar() {
+      const userImage = Backend.getProfileImageUrl(this.address);
+      return !this.address ? this.defaultAvatar : userImage;
```
venimus, line 35: Shouldn't `this.defaultAvatar` be somehow replaced by identicon()

src/components/Avatar.vue
mradkov: We check if the user has uploaded custom avatar anyways by making a request to the backend. 404 is the error code which is returned if none was found.

venimus: Imho this should be `v-if="!avatar"`

src/components/Avatar.vue
venimus: I would just return null, and check that in the template, where it switches to rendering the .svg result

src/components/Avatar.vue
mradkov: `@error` is a correct way of handling it IMO as it executes a state change in the component and renders the icon if the backend returns 404.

its not possible to use it as reactive property as the request is async.

src/views/UserProfileView.vue
venimus: Imho the style might be better in the parent component, which can set the effects /sizes and etc required in its context, so I think scoped should probably stay, tough i don't seem much problem as the styling in this case is only for a component that appears only 1 on the page (also the style selectors are css-scoped already - not too generic)

src/components/Avatar.vue
venimus: Might be having "avatar" now taking null will not trigger a new request but instantly switch to the identicon, which is acceptable for me. Tough still is hackish.

venimus: You could also add  `loading="lazy"` to the img tag to save some resources

src/components/Avatar.vue
```diff
@@ -0,0 +1,48 @@
+<!-- eslint-disable-next-line vue/no-v-html -->
```
venimus, line 1: i guess this should be put right above the `img` tag or disabled in file (now disable-next-line points to `<template>`).

src/views/UserProfileView.vue
stanislav-slavov: I saw that class .description in User Profile Page is breaking the description in preview of a tip. If we are removing "scoped", please add wrapper class for those pages. For example all styles for User Profile page should go under .profile__page .

PR #324, opened 2020-04-09
Created single comment view 

stanislav-slavov:
ref #321
ref #445
ref #446


mradkov:
Screenshot 2020-04-10 at 17 15 57

The comment ID in the URL is mistaken. Url should be: `/tip/TIP_ID/comment/COMMENT_ID`

src/router.js
```diff
@@ -17,13 +18,33 @@ const guardTipComments = (to, from, next) => {
...
+
 const routes = [
   {
     path: '/',
     name: 'home',
     component: TipsList,
     meta: { title: 'Home' },
   },
+  {
+    path: '/tip/:id/comment/:id',
```
mradkov, line 39:
```suggestion
    path: '/tip/:tipId/comment/:id',
```

venimus: can we request an API entrypoint for comments? this does not seem logical if comments has unique ID, you don't need the tipId to request the entity data

mradkov: I think we should keep the routes like this, for easier filtering later on in the comment section and also easier associating on-chain and off-chain comments with the specific tip.

venimus: Also I think this should be done with `route children`

src/views/SingleCommentView.vue
```diff
@@ -0,0 +1,184 @@
...
+  },
+  beforeDestroy() {
+    clearInterval(this.interval);
+  },
+  methods: {
+    getAvatar(address) {
+      const userImage = Backend.getProfileImageUrl(address);
+      return userImage || this.avatar;
+    },
+    // async sendTipComment() {
```
venimus, line 140: can you clean up these, or it is still WIP

src/router.js
mradkov: https://github.com/aeternity/tipping-community-backend/blob/superhero/routes/README.md

src/views/SingleCommentView.vue
stanislav-slavov: This should be implemented - sending comment to another comment

thepiwo:
lets close this for now as there hasn't been any progress the last week, feel free to repopen

mradkov:
The backend now supports the nested comments. Lets revive this.

davidyuk:
The support of nested comments implemented in the wallet in this PR: https://github.com/aeternity/superhero-wallet/pull/259

mradkov:
Screenshot 2020-05-05 at 0 17 17

mradkov:
Can we also visualize the reply better as it is in the designs:

Screenshot 2020-05-05 at 0 18 35

src/router.js
```diff
@@ -17,13 +18,33 @@ const guardTipComments = (to, from, next) => {
   }
 };
 
+const guardCommentView = (to, from, next) => {
+  if (to.name === 'comment'
+      && typeof to.params.id !== 'undefined'
+      && typeof to.params.tipId !== 'undefined') {
+    next();
+  } else {
+    next('/');
```
venimus, line 27: why not 404?

src/components/tipRecords/TipComment.vue
```diff
@@ -1,39 +1,45 @@
...
-            <Avatar :address="comment.author" />
-            <span
-              v-if="userChainName"
-              class="chain__name"
+          <span
+            class="user-display"
+            @click.stop
+          >
+            <router-link
+              :to="'/user-profile/' + comment.author"
```
venimus, line 19: IMHO best is to pass params as object

src/router.js
stanislav-slavov: We don't have 404 page. I guess this a matter of a different user-story. Our maintenance page is there to show that the site is down.

Issue #323, opened 2020-04-09
Remove Trending from main navigation


venimus:
It is doing nothing and takes space. Could be reintroduced when its functionality is ready. Don't delete the icons

PR #325, opened 2020-04-09
Removed "trending" link from main navigation

stanislav-slavov:
ref #323
OP task: #693

### DIFF
--- a/src/components/layout/RightSection.vue
+++ b/src/components/layout/RightSection.vue
@@ -108,7 +108,7 @@ export default {
     currencyDropdownOptions() {
       return Object.keys(this.currencyRates.aeternity)
         .map((key) => ({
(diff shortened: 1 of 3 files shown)


src/components/layout/RightSection.vue
```diff
@@ -108,7 +108,7 @@ export default {
     currencyDropdownOptions() {
       return Object.keys(this.currencyRates.aeternity)
         .map((key) => ({
-          text: `${this.getFiatVal(this.currencyRates.aeternity[key])} ${key.toUpperCase()}`,
+          text: `~${this.getFiatVal(this.currencyRates.aeternity[key])} ${key.toUpperCase()}`,
```
venimus, line 111: I think this ~ is very annoying we should probably remove it from everywhere. Can you revert that change. We have to discuss it first probably.

PR #326, opened 2020-04-09
Created back button to use browser history

stanislav-slavov:
ref #327
OP task: #700

### DIFF
--- a/src/components/BackButtonRibbon.vue
+++ b/src/components/BackButtonRibbon.vue
@@ -1,0 +1,45 @@
+<template>
+  <div class="actions-ribbon">
+    <img
+      src="../assets/backArrow.svg"
+      @click="$router.go(-1)"
+    >
+  </div>
+</template>
+
+<script>
+
+export default {
+  name: 'BackButtonRibbon',
+};
+</script>
+
+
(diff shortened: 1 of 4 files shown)


src/components/BackButtonRibbon.vue
```diff
@@ -0,0 +1,45 @@
...
+  <div class="actions-ribbon">
+    <img
+      src="../assets/backArrow.svg"
+      @click="$router.go(-1)"
+    >
+  </div>
+</template>
+
+<script>
+
```
venimus, line 11:
```suggestion
```

src/components/BackButtonRibbon.vue
```diff
@@ -0,0 +1,45 @@
...
+export default {
+  name: 'BackButtonRibbon',
+};
+</script>
+
```
venimus, line 16:
```suggestion
```

venimus:
It's good to merge, but it does not really solve the original problem because the pages still reloads after going back and scroll position is lost. May be we need to think of something else sometimes later

venimus:
I made another PR that should complete the feature

PR #330, opened 2020-04-09
Fix Comments view background

venimus:
follow up of #319


mradkov:
Screenshot 2020-04-10 at 7 48 44

Does not look good. The comments background is the same as the feed

venimus:
> Does not look good. The comments background is the same as the feed

@mradkov can you recheck

PR #331, opened 2020-04-10
Fixed tip preview padding on profile page

stanislav-slavov:
ref #332
OP bug: #705


src/views/UserProfileView.vue
```diff
@@ -673,18 +686,6 @@ export default {
     }
   }
```
venimus, line 688:
```suggestion
```

PR #335, opened 2020-04-10
Created tip-control component

stanislav-slavov:


src/components/tipRecords/TipRecord.vue
```diff
@@ -34,8 +34,8 @@
         class="tip__article"
       >
         <a
-          :href="tip.url"
           target="_blank"
+          @click="goToUrl(tip.url)"
```
venimus, line 38: why is that needed?
This way you can't see the link tooltip of the browser, also the `<a>` tag becomes no-different than a `div`

stanislav-slavov: Because we prevent the click action from bubbling, otherwise the link gets its href executed and always open a new tab. In this case we want to open the dialog

src/components/TipControl.vue
```diff
@@ -0,0 +1,205 @@
...
+          class="input-group"
+        >
+          <input
+            v-model="message"
+            type="text"
+            class="form-control tip__message"
+            placeholder="Add message"
+          >
+        </div>
+        <div v-if="!showLoading">
```
venimus, line 53: I would have made that a `<form>` with `@onsubmit="submitAction()` and put the `<ae-button>` to render a `button type=submit`. This way pressing Enter will work also will be somehow better supported on mobile

src/components/TipControl.vue
```diff
@@ -0,0 +1,205 @@
...
+      USE_DEEP_LINKS,
+      message: '',
+    };
+  },
+  computed: {
+    ...mapGetters(['account', 'balance', 'isLoggedIn', 'loading']),
+    eventPayload() {
+      return `${this.tip.id}:${this.show}`;
+    },
+    // deepLink() {
```
venimus, line 110: if it is not used, better not leave "commented probably-features"

mradkov: The deeplink should be used. Maybe @davidyuk can take a look at the flow and see if it needs adjustment.

src/components/tipRecords/TipRecord.vue
venimus: It should not happen if you use `@click.stop="handler"`

src/components/TipControl.vue
venimus: so better tag it WIP

src/components/TipControl.vue
```diff
@@ -154,14 +172,13 @@ export default {
     async sendTip() {
       const amount = util.aeToAtoms(this.value);
       this.showLoading = true;
-      await aeternity.tip(this.tip.url, this.title, amount)
+      await aeternity.tip(`${window.location.origin}/#/tip/${this.tip.id}`, this.message, amount)
```
venimus, line 175: can we first consider to use url without the domain. I'm very concerned about leaking random internal urls stored in the blockchain.

venimus: also it makes this untestable

mradkov: URL without the domain is not URL, rather a path.

src/components/AeButton.vue
```diff
@@ -2,6 +2,7 @@
   <button
     :disabled="disabled"
     class="btn btn-primary ae-button"
+    type="submit"
     @click="$emit('click', $event)"
```
venimus, line 6: is this something functional?

src/components/TipControl.vue
```diff
@@ -0,0 +1,234 @@
...
+        <loading :show-loading="showLoading" />
+        <div
+          v-show="error && !showLoading"
+          class="text-center mb-2"
+        >
+          An error occurred while sending tip
+        </div>
+        <form
+          v-if="!showLoading"
+          @submit.prevent
```
venimus, line 62: doesn't this work `@submit.prevent="submitAction()"` (and remove it from the button)? Enter on the focused input emits submit event to the form, so the form should handle it. The same way button=submit emits the same event to the form, so we usually handle the event there. Then other onclick stuff become useless (including the one in ae-button). If `<ae-button>` is to be used as generic button, so we better make type to be configurable (as component property)

src/components/AeInputAmount.vue
```diff
@@ -11,7 +11,7 @@
       aria-label="Default"
       aria-describedby="inputGroup-sizing-mn"
       @input="$emit('input', $event.target.value)"
```
venimus, line 13: Not sure about these either

src/components/AeButton.vue
```diff
@@ -2,6 +2,7 @@
   <button
     :disabled="disabled"
     class="btn btn-primary ae-button"
+    type="submit"
```
venimus, line 5: I think its best this to be passed as a component prop

src/components/Retip.vue
```diff
@@ -170,33 +170,9 @@ export default {
...
-  .ae-amount {
-    align-items: center;
-    display: flex;
-    height: 1rem;
-  }
-
-  .retip__content {
+<style lang="scss">
+  .retip__content,
+  .tip__content {
```
venimus, line 175: move all tip__content related css to the other component

src/components/TipControl.vue
```diff
@@ -0,0 +1,234 @@
...
+          >
+            <input
+              v-model="message"
+              type="text"
+              class="form-control tip__message"
+              placeholder="Add message"
+            >
+          </div>
+          <div class="amount__row">
+            <ae-input-amount v-model="value" />
```
venimus, line 75:
```suggestion
            <ae-input-amount v-model.number="value" />
```

venimus: same should be to retip

venimus: Else, there is a validation issue: when you edit the number field and return it to 0, it becomes '0' (as string) and validation starts failing

PR #345, opened 2020-04-13
Feature card embedly previews

mradkov:
Embedly previews.

Check also: https://meta.discourse.org/tag/onebox


mradkov:
This is blocked by problem with embedly api.

mradkov:
Check also: https://meta.discourse.org/tag/onebox

mradkov:
When you enter into single tip preview of such tip and then go back - the preview gets broken. cc @petbaik

petbaik:
> When you enter into single tip preview of such tip and then go back - the preview gets broken. cc @petbaik

Since the router view is wrapped in keep-alive element (https://github.com/aeternity/superhero-ui/blob/develop/src/App.vue#L16) components are cached. The problem is that embdely previews are iframes and looks like keep-alive cannot cache iframe and when enter into single tip the `iframe` element is removed from the DOM. When go back components are restored but the iframes are missing.

mradkov:
> > When you enter into single tip preview of such tip and then go back - the preview gets broken. cc @petbaik
> 
> Since the router view is wrapped in keep-alive element (https://github.com/aeternity/superhero-ui/blob/develop/src/App.vue#L16) components are cached. The problem is that embdely previews are iframes and looks like keep-alive cannot cache iframe and when enter into single tip the `iframe` element is removed from the DOM. When go back components are restored but the iframes are missing.

Can we force reload them via custom event in the EventBus or maybe if you have some other suggestion?

thepiwo:
The code seems like this is an approach with lots of overhead

venimus:
We can use keep-alives only for real browsers (it is to provide proper UI navigation for users). So if we can check if it is a headless browser just not use the keep-alive tag (set cache-ount to 0?)

mradkov:
> We can modify to use keep-alives only for "real" browsers (it is there only to provide proper UI navigation for users). So if we can check if it is a headless browser just not use the keep-alive tag (set cache-ount to 0?)

This feature has nothing to do with headless browsers.

mradkov:
@kenodressel Can we generate the embedly previews in the backend?

kenodressel:
I guess we could but it looks like the API is only available in the larger package which is starting at 99$ a month.

Issue #350, opened 2020-04-14
Visual issues with tip previews (tip popup forms)


venimus:
![image](https://user-images.githubusercontent.com/632282/79276454-b426b180-7eb0-11ea-8ba0-148870049323.png)

- [x] Inner icon (diamond) should be larger (this i tracked is caused by a messy css)
- [x] Input color should be white on both
- [x] z-index of popups is messy (text appear above)
- [x] Should open only one tip popup at a time (again cause by messy `.overlay` z-index, but also the div should be moved in DOM)

PR #352, opened 2020-04-15
Remove Retip component, extract AeAmountFiat component

### LINKED ISSUE #342: Cleanup css of retip and tip components
Currently retip component CSS is messing with the TipComponent (or use the same classes) should be separated, the preview tip icon should be fixed

CedrikNikita:
closes #342 
part of #350
Fix inconsistency between
| Main view |  Tipview |
|---|---|
| image | image|


src/components/TipInput.vue
```diff
@@ -125,22 +112,22 @@ export default {
...
   },
   computed: {
-    ...mapGetters(['account', 'balance', 'isLoggedIn', 'loading']),
+    ...mapGetters(['account', 'loading']),
     eventPayload() {
       return `${this.tip.id}:${this.show}`;
     },
     deepLink() {
-      const url = new URL(`${process.env.VUE_APP_WALLET_URL}/tip`);
+      const url = new URL(`${process.env.VUE_APP_WALLET_URL}/${this.isRetip ? 're' : ''}tip`);
```
venimus, line 120: This kind of unreadable, I would put that in a separate method that returns the url (endpoint)

src/components/TipInput.vue
```diff
@@ -125,22 +112,22 @@ export default {
...
       url.searchParams.set('id', this.tip.id);
       url.searchParams.set('x-success', window.location);
       url.searchParams.set('x-cancel', window.location);
       return url;
     },
     isSendMessageDataValid() {
-      return this.message.trim().length > 0 && !this.value;
```
venimus, line 140: there was some logic behind this (tough not very nice implemented):
You either make a tip + on-chain comment (needs tip value >0 and a message) or make an off-chain comment (where tip value is 0)

src/components/TipInput.vue
```diff
@@ -150,6 +137,17 @@ export default {
...
       return this.isTipped ? iconTipped : iconTip;
     },
+    amount() {
+      return this.isRetip
+        ? this.tip.total_amount
+        : this.tip.retip_amount_ae;
+    },
+    title() {
+      if (this.isRetip) {
+        return this.isTipped ? 'Total tips (you tipped too)' : 'Total tips (click to retip the same URL)';
```
venimus, line 147: Though this is wokring, IMHO all text data should be better in the template (as it could be easily translated, e.g. with filters)

src/components/TipInput.vue
mradkov: > there was some logic behind this (tough not very nice implemented):
> You either make a tip + on-chain comment (needs tip value >0 and a message) or make an off-chain comment (where tip value is 0).
> Third option is "retip" where you don't really add a message AFAIK
> May be double check that (@mradkov)?

Yes, correct.

src/components/TipInput.vue
davidyuk:
```suggestion
      const url = new URL(`${process.env.VUE_APP_WALLET_URL}/${this.isRetip ? 'retip' : 'tip'}`);
```
Would you prefer this more? Actually, this is already a method (computed property) that returns the url, don't see a reason to split it into two methods.

src/components/TipInput.vue
davidyuk: Does `vue-i18n` support translation based on filters? I don't see a significant difference to have texts here or in markup, it is more important to don't have to lot of business logic in components by extracting it to vuex store (at least this will make unit testing easier). To translate something in a template we are usually using `{{ t('translation-key') }}`, to do the same in the js section of a component we can use `this.$t('translation-key')`, so it shouldn't be more difficult to translate something in js section.

PR #354, opened 2020-04-15
Added texts, fix typos

### LINKED ISSUE #353: Update wizzard, onboarding, FAQ and tutorial texts
Some typos reported.

mradkov:
This closes #353

### DIFF
--- a/src/views/CreateProfile.vue
+++ b/src/views/CreateProfile.vue
@@ -6,13 +6,15 @@
     <div class="container wrapper">
       <div class="content tutorial-page">
         <h1>
-          Join Superhero League: <br>
+          HOW TO USE SUPERHERO.COM: <br>
           fully P2P tipping with no boundaries,<br>
           no censorship - at near-zero fees
         </h1>
         <hr width="30%">
         <expandable-block :title="'HOW TO USE SUPERHERO'">
-          <p><b>Download and install the Superhero extension for your browser.</b></p><p /><p>
+          <p><b>Download and install the Superhero extension for your browser.</b></p>
+          <p />
+          <p>
             Superhero is currently available for Firefox and Chrome.<br>
             (The Chrome extension can be used for Brave and Opera.)
           </p>
@@ -23,14 +25,12 @@
               href="https://addons.mozilla.org/en-US/firefox/addon/superhero-wallet/"
               target="_blank"
             >
-              here
-            </a><br>
+              here </a><br>
             Click
             <span class="highlited">Add</span>
             to Firefox. Confirm by clicking
             <span class="highlited">Add</span> again in the message prompt.<br>
-            Then the Superhero icon should appear on the upper right
-            corner of your Firefox browser
+            Then the Superhero icon should appear on the upper right corner of your Firefox browser
             tool bar.<br>
           </p>
           <p>
@@ -39,12 +39,10 @@
             <a
               :href="chromeLink"
               target="_blank"
-            >
-              here
-            </a><br>
+            > here </a><br>
             Add the extension to your browser.<br>
-            Then the Superhero icon should appear on the
-            upper right corner of your Chrome browser tool bar.
+            Then the Superhero icon should appear on the upper right corner of your Chrome browser
+            tool bar.
           </p>
         </expandable-block>
         <expandable-block :title="'CREATING YOUR WALLET'">
@@ -57,26 +55,24 @@
           <p>
             You can find your wallet address on your dashboard, right above your balance.<br>
             It’s a random string of numbers and letters that starts with ak_.<br>
-            This wallet address is also referred to as a public
-            key— it’s what others see as your public ID by default.<br>
-            You can easily copy this by clicking the copy
-            button right next to your wallet address.
+            This wallet address is also referred to as a public key— it’s what others see as your
+            public ID by default.<br>
+            You can easily copy this by clicking the copy button right next to your wallet address.
           </p>
         </expandable-block>
         <expandable-block :title="'SECURING YOUR WALLET'">
           <p>
             With great freedom comes great responsibility.<br>
             You’re in charge of keeping the keys to your account.<br>
-            There are no third-party custodians to store
-            nor connect any password with any account.<br>
+            There are no third-party custodians to store nor connect any password with any
+            account.<br>
             The task of keeping the keys to your wallet is yours and yours alone, Superhero.
           </p>
           <p>
             Here’s how to secure and backup your wallet:<br>
-            Once you’re in your dashboard, you’ll get a series
-            of words called a seed phrase.<br>
-            The seed phrase is somewhat like a password –
-            it’s a unique set of words (in order) which connects you to your wallet.<br>
+            Once you’re in your dashboard, you’ll get a series of words called a seed phrase.<br>
+            The seed phrase is somewhat like a password – it’s a unique set of words (in order)
+            which connects you to your wallet.<br>
           </p>
           <ul>
             <li>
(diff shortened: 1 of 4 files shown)


src/views/CreateProfile.vue
```diff
@@ -39,12 +39,10 @@
...
               :href="chromeLink"
               target="_blank"
-            >
-              here
-            </a><br>
+            > here </a><br>
             Add the extension to your browser.<br>
-            Then the Superhero icon should appear on the
-            upper right corner of your Chrome browser tool bar.
+            Then the Superhero icon should appear on the upper right corner of your Chrome browser
```
venimus, line 44:
```suggestion
            Then the Superhero icon should appear on the upper right corner of your browser
```

src/views/CreateProfile.vue
```diff
@@ -39,12 +39,10 @@
...
+            tool bar.
```
venimus, line 45:
```suggestion
            Then the Superhero icon should appear on the upper right corner of your browser
            toolbar.
```

src/views/CreateProfile.vue
```diff
@@ -91,14 +87,13 @@
...
+            You’ll need this seed phrase to retrieve your wallet in case you lose access to the
+            extension for any reason (migrating to a new device, reinstalling browser, etc.)
           </p>
           <ul>
             <li>
-              <b>Write down the 12 words displayed on a piece of
-                paper to keep in a safe place only you can find.
+              <b>Write down the 12 words displayed on a piece of paper to keep in a safe place only
+                you can find.
               </b>
```
venimus, line 97:
```suggestion
              <b>Write down the 12 words displayed on a piece of paper and keep it in a safe and secret place.</b>
```

src/views/CreateProfile.vue
```diff
@@ -113,45 +108,35 @@
...
-              Open your Superhero wallet extension.
-              The button <img src="../assets/iconExtension.svg">
-              should be in the
-              upper right of your browser tool bar.<br>
-            </li>
-            <li>
-              Click <span class="highlited">Tip</span> someone.
+              Open your Superhero wallet extension. The button
+              <img src="../assets/iconExtension.svg"> should be in the upper right of your browser
+              tool bar.<br>
```
venimus, line 113:
```suggestion
              toolbar.<br>
```

src/views/CreateProfile.vue
```diff
@@ -113,45 +108,35 @@
...
             </li>
+            <li>Click <span class="highlited">Tip</span> someone.</li>
```
venimus, line 113:
```suggestion
            <li>Click <span class="highlited">Send Tip</span></li>
```

src/views/CreateProfile.vue
```diff
@@ -113,45 +108,35 @@
...
-              On the next screen, fill in the required fields.
-              Enter the amount of tips you want to send, and add a
-              message telling the receiver why they’re awesome.
-            </li>
-            <li>
-              Click <span class="highlited">Send</span> tip.
+              On the next screen, fill in the required fields. Enter the amount of tips you want to
+              send, and add a message telling the receiver why they’re awesome.
             </li>
+            <li>Click <span class="highlited">Send</span> tip.</li>
```
venimus, line 118:
```suggestion
            <li>Click the <span class="highlited">Send Tip</span> button below.</li>
```

src/views/CreateProfile.vue
```diff
@@ -193,30 +178,28 @@
...
             >
-              here
-            </a>.
+              here </a>.
           </p>
           <p>
-            You can claim as many names on Superhero as you want.
-            But as a rule, you must have a minimum of XX AE in your wallet.
+            You can claim as many names on Superhero as you want. But as a rule, you must have a
+            minimum of XX AE in your wallet.
```
venimus, line 200: XX ? should be specified

src/views/FAQ.vue
```diff
@@ -11,13 +11,13 @@
           if you can't find the answer of you question below visit our
           <router-link to="/create-profile">
             Tutorial Page
-          </router-link>to
-          <br>receive more information about using Superhero platform and wallet.
+          </router-link>to <br>receive more
+          information about using Superhero platform and wallet.
```
venimus, line 15:
```suggestion
          </router-link> to<br>
          receive more information about using Superhero platform and wallet.
```

src/views/FAQ.vue
```diff
@@ -32,33 +32,29 @@
...
-          <br>No 3
-          <sup>rd</sup> party custodians will control
-          content, or your ability to tip or receive tips.
-          <br>Superhero is you and you and you and you… Without any “them” in-between.
+          >aeternity blockchain</a>. <br>Anyone,
+          anywhere may use the platform. <br>Other websites may use the platform. <br>All have
+          uncensored, equal ability to be Superhero. <br>As P2P – the only ones who are not part
+          of Superhero intermediaries. <br>No 3 <sup>rd</sup> party custodians will control
+          content, or your ability to tip or receive tips. <br>Superhero is you and you and you
+          and you… Without any “them” in-between.
```
venimus, line 40: the linter requires `<br>` to be the last on the line with a break after it (AFAIK)
```suggestion
          >aeternity blockchain</a>.<br>
          Anyone, anywhere may use the platform.<br>
          Other websites may use the platform.<br>
          All have uncensored, equal ability to be Superhero.<br>
          As P2P – the only ones who are not part of Superhero intermediaries.<br>
          No 3 <sup>rd</sup> party custodians will control content,
          or your ability to tip or receive tips.<br>
          Superhero is you and you and you and you… Without any “them” in-between.
```

src/views/FAQ.vue
```diff
@@ -32,33 +32,29 @@
...
         </expandable-block>
         <expandable-block :title="'How does Superhero work?'">
           Superhero is built on aeternity blockchain.
-          <br>Because it is built on blockchain, actions are transparent.
-          <br>Because it is built on the blockchain, actions are published forever.
-          <br>Blockchain has promised a more free and decentralized future for society.
-          <br>Superhero used blockchain as its foundation to make that promise a reality.
-          <br>Superhero is an online community making a tangible difference in the real world.
+          <br>Because it is built on blockchain, actions are transparent. <br>Because it is
```
venimus, line 44: the linter requires `<br>` to be the last on the line with a break after it (AFAIK)

venimus: Tough not a problem for me per se

mradkov: > the linter requires `<br>` to be the last on the line with a break after it (AFAIK)

Strange, because I run it before every commit and it does not complain.

src/views/CreateProfile.vue
mradkov: Good question. I guess we should not limit this to specific amount as it will eventually depend on the length of the name that the user want to claim.

src/views/FAQ.vue
venimus: I think it is a warning which fails on travis, but might not fail locally

venimus: But if it passes then just don't change it. May be i'm mistaken

src/views/CreateProfile.vue
venimus: I though about that, currently from what I understood you need to have at least 12 chars to claim from within the wallet which requires 2-3AE for it (or something like that) I think that might be the minimum we can put here. You will require more for shorter but you functionally need to use the auctions, which is not covered here. May be we can just mention that, so ppl are not confused

mradkov: I thought of just slightly modifying the text to not point to an exact value.

"... you must have enough AE to cover the price for the name ... "

Issue #364, opened 2020-04-21
Tip/Retip functionalities not working


stanislav-slavov:
The validation passes but the action never executes.

PR #365, opened 2020-04-21
Execute correct action upon validation

stanislav-slavov:
ref #364

### DIFF
--- a/src/components/TipInput.vue
+++ b/src/components/TipInput.vue
@@ -162,7 +162,10 @@ export default {
     submitAction() {
       if (!this.isRetip && this.isMessageValid && !this.value) {
         this.sendTipComment();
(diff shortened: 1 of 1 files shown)


src/components/TipInput.vue
```diff
@@ -158,9 +160,9 @@ export default {
   },
   methods: {
     submitAction() {
-      if (!this.isRetip && this.isSendMessageDataValid) {
+      if (!this.isRetip && this.isMessageValid && !this.value) {
         this.sendTipComment();
-      } else if (this.isSendTipDataValid) {
+      } else if (this.isDataValid) {
```
venimus, line 165:
```suggestion
        return;
      }

      if (this.isDataValid) {
```

Issue #368, opened 2020-04-21
Retip action


mradkov:
<img width="647" alt="Screenshot 2020-04-21 at 20 58 31" src="https://user-images.githubusercontent.com/13139371/79898050-44b24400-8413-11ea-995d-56844a59c0d3.png">

1. Should retip the URL from the preview.
2. Should retip the internal single-tip-preview-url (/tip/tip_id)

This is still not the case I think

PR #370, opened 2020-04-22
Tip amounts are retrived from stats

stanislav-slavov:
ref #368

### DIFF
--- a/src/components/TipInput.vue
+++ b/src/components/TipInput.vue
@@ -140,7 +140,7 @@ export default {
     },
     derivedTipStats() {
       if (!this.stats || !this.stats.by_url) {
-        return 0;
+        return null;
       }
       if (this.isRetip) {
         return this.stats.by_url.find((tipStats) => tipStats.url === this.tip.url);
--- a/src/components/TipInput.vue
+++ b/src/components/TipInput.vue
@@ -145,7 +145,7 @@ export default {
       if (this.isRetip) {
         return this.stats.by_url.find((tipStats) => tipStats.url === this.tip.url);
       }
-      return this.stats.by_url.find((tipStats) => tipStats.url === `${window.location.origin}/#/tip/${this.tip.id}`);
+      return this.stats.by_url.find((tipStats) => tipStats.url === `https://beta.superhero.com/#/tip/${this.tip.id}`);
     },
     amount() {
       return this.derivedTipStats ? this.derivedTipStats.total_amount : '0';
--- a/src/components/TipInput.vue
+++ b/src/components/TipInput.vue
@@ -112,7 +112,7 @@ export default {
     };
   },
   computed: {
-    ...mapGetters(['account', 'loading']),
+    ...mapGetters(['account', 'loading', 'stats']),
     eventPayload() {
       return `${this.tip.id}:${this.show}`;
     },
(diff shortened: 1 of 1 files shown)


src/components/TipInput.vue
```diff
@@ -130,25 +130,33 @@ export default {
...
       return this.isTipped ? iconTipped : iconTip;
     },
+    derivedTipStats() {
+      if (!this.stats || !this.stats.by_url) {
+        return false;
+      }
+      if (this.isRetip) {
+        return this.stats.by_url.find((tipStats) => tipStats.url === this.tip.url);
+      }
+      return this.stats.by_url.find((tipStats) => tipStats.url === `${window.location.origin}/#/tip/${this.tip.id}`);
```
thepiwo, line 148: should probably use a static url for superhero, as otherwise the content breaks if the url changes

src/components/TipInput.vue
```diff
@@ -130,25 +130,33 @@ export default {
...
-        || this.tip.retips.filter((retip) => retip.sender === this.account).length > 0;
+      if (!this.stats || !this.derivedTipStats) {
+        return false;
+      }
+      return this.derivedTipStats.senders.find((sender) => sender === this.account);
     },
     iconTip() {
       return this.isTipped ? iconTipped : iconTip;
     },
+    derivedTipStats() {
```
thepiwo, line 141: maybe this should just return derivedTipAmount  and 0 if something is not found.

src/components/TipInput.vue
stanislav-slavov: https://beta.superhero.com/#/tip/${this.tip.id} is this what you mean?
I guess it will be harder to test it.

thepiwo: probably the final final url or an abstract representation @mradkov

mradkov: I thought of having a prefix or custom schema: e.g. `superhero://tip/123`

But this wont be an correct URL then and oracles would require tweaking.

thepiwo: we are in control what we open? so superhero:// can be adjusted to local environment before opening

thepiwo: oracles need adjustment anyways in order to properly support this feature

mradkov: I suggest we use `beta.superhero.com` for now. When sh is moved to the final URL we can adjust it. (still tips on beta will be able to be claimed).

mradkov, line 148:
```suggestion
      return this.stats.by_url.find((tipStats) => tipStats.url === `https://beta.superhero.com/#/tip/${this.tip.id}`);
```

src/components/TipInput.vue
```diff
@@ -130,25 +130,33 @@ export default {
...
+      if (!this.stats || !this.stats.by_url) {
+        return 0;
```
venimus, line 143: this should probably return `null` as it is compatible with the other return results (which return object or null)

src/components/TipInput.vue
```diff
@@ -130,25 +130,33 @@ export default {
       return (this.value > 0 && (this.isRetip || this.isMessageValid)) || this.isMessageValid;
     },
     isTipped() {
-      return !this.loading
-        || (this.tip.sender === this.account)
-        || this.tip.retips.filter((retip) => retip.sender === this.account).length > 0;
+      if (!this.stats || !this.derivedTipStats) {
+        return false;
+      }
+      return this.derivedTipStats.senders.find((sender) => sender === this.account);
```
venimus, line 136: this should return boolean 
```suggestion
      return this.derivedTipStats.senders.find((sender) => sender === this.account) !== null;
```
or hackish
```suggestion
      return !!this.derivedTipStats.senders.find((sender) => sender === this.account);
```

mradkov:
Putting this on hold until its clear how to proceed with these buttons and we have the accompanied filters in place to handle the new flow.

What's your opinion? cc @venimus

venimus:
> Putting this on hold until its clear how to proceed with these buttons and we have the accompanied filters in place to handle the new flow.
> 
> What's your opinion? cc @venimus

Agree, seems contract migration is on hold so does tasks which rely on it

PR #400, opened 2020-04-27
Bugfix/rename home page

stanislav-slavov:
ref #399


stanislav-slavov:
The base of this branch is not from develop

Issue #398, opened 2020-04-27
display tip form all the time


mradkov:
Display tip form at all the time. If the user does not have wallet installed - display notification they need one in order to tip.

venimus:
May be we can always show a yellow message that in order to tip they need to have the wallet and logged in it.

PR #404, opened 2020-04-27
Show Send Tip form at any time

stanislav-slavov:
ref #398

### DIFF
--- a/src/components/layout/SendTip.vue
+++ b/src/components/layout/SendTip.vue
@@ -41,7 +41,7 @@
         v-if="!canTip"
         class="install-wallet-warning"
       >
-        The wallet should be installed and active in order to tip.
+        You need to have a wallet installed and active in order to tip.
       </div>
       <div class="text-right">
         <ae-button
--- a/src/components/layout/SendTip.vue
+++ b/src/components/layout/SendTip.vue
@@ -207,7 +207,7 @@ export default {
     .install-wallet-warning {
       text-align: right;
       font-size: 0.75rem;
-      color: $light_font_color;
+      color: $warning_font_color;
       margin-bottom: 0.5rem;
     }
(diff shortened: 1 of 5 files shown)


stanislav-slavov:
I have added an orange warning message. 
The whole form is disabled and the warning is shown if there is no wallet or the user is not logged.
I don't think we should display the warning if everything is ok with the account.

src/styles/_variables.scss
```diff
@@ -15,11 +15,10 @@ $thumbnail_background_color_alt: #33343e;
...
 
-//Custom styles
 $custom_links_color: #67f7b8;
 $search_icon_hover_color: $white_color;
 $search_icon_backgound: #52528b;
-$article_description_color: #96969b;
 $comment_text_color: #8f9095;
 $tip_note_color: #babac0;
 $text_content_color: #d2d2d4;
+$warning_font_color: #856404;
```
venimus, line 24:
```suggestion
$warning_font_color: #ffdc50;
```
We need to use only design consistent colors. This one is approved for warning tough it is not yet implemented elsewhere. It is for "site badges" - warning, we can use it sitewise

stanislav-slavov: Thanks for providing it.

src/components/layout/SendTip.vue
```diff
@@ -27,15 +27,25 @@
...
+            v-model="sendTipForm.amount"
+            :disabled="!canTip"
+          />
         </div>
       </div>
+      <div
+        v-if="!canTip"
+        class="install-wallet-warning"
+      >
+        The wallet should be installed and active in order to tip.
```
mradkov, line 44:
```suggestion
        You need to have a wallet installed and active in order to tip.
```

Issue #403, opened 2020-04-27
Remove/Replace Create Profile


venimus:
Contents should be rephrased as FAQ sections
Menu navigation should not appear for not logged in users or use something less confusing like Install wallet tutorial

mradkov:
Create profile should not be removed, rather lead to installing wallet or ?

venimus:
May be we can create (rename it to) a "Install the wallet" page or something and put some info there for how to do that (extract text from FAQ and style it a little)

stanislav-slavov:
Create-Profile page is styled exactly as the FAQ page. Also has the same expandable panels with informative sections. As far as I understand the requirement - we should rename the Create Profile link to "Install the wallet"

PR #408, opened 2020-04-27
Renamed Create Profile to Install Wallet

stanislav-slavov:
ref #403

### DIFF
--- a/src/components/layout/Navigation.vue
+++ b/src/components/layout/Navigation.vue
@@ -38,9 +38,9 @@
       class="navigation__item profile"
       @click="scrollTop()"
     >
-      <router-link :to="{ name: 'install-wallet' }">
+      <router-link :to="{ name: 'tutorial' }">
         <div class="navigation__item__image" />
-        <span>Install Wallet</span>
+        <span>Tutorial</span>
       </router-link>
     </div>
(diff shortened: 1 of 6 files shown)


src/components/layout/Navigation.vue
```diff
@@ -38,9 +38,9 @@
       class="navigation__item profile"
       @click="scrollTop()"
     >
-      <router-link :to="{ name: 'create-profile' }">
+      <router-link :to="{ name: 'install-wallet' }">
         <div class="navigation__item__image" />
-        <span>Create Profile</span>
+        <span>Install Wallet</span>
```
venimus, line 43: May be it should be 'Install a wallet' - it is grammatically correct

venimus: However the content is kind of a mess there. It says How to Use superhero
So probably we may just rename it Tutorial

venimus:
It appears that the request is to completely remove this page. There will be a Tutorial elsewhere @mradkov did I got it right?

mradkov:
> It appears that the request is to completely remove this page. There will be a Tutorial elsewhere @mradkov did I got it right?

Remove it from the navigation then. Keep the content we can move it in the FAQ in another PR.

Issue #297, opened 2020-04-06
Not supported browser


mradkov:
Display warning when user opens superhero from a browser which we don't support (do not have extension available for). E.g. safari, IE.

venimus:
I really think we have to add few of the chrome-based browsers that can install the extension to "supported" also Opera, Vivaldi are already mentioned in Help sections

venimus:
I reopen the issue, because there will be some changes in the Wizard that mention other browsers (TBD), and will be awkward to show Unsupported on top, while the wizard explains the opposite. So we can update the list later. 
Also I'm also concerned about this feature in general, as it is not really clear what "Unsupported" means. Site works on most browsers, but the extension does not. Yet using the site just for looking around does not require you to have an extension. Furthermore we have that backup web wallet, which does not require extension and people still can use the site's full functionality.
 So IMHO this message should be somehow in the context of wallet support/installation, not the aggregator in general.

PR #415, opened 2020-04-28
Update browser list that support the extension

venimus:
close https://github.com/aeternity/superhero-ui/issues/297

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -11,7 +11,8 @@
         v-if="isSupportedBrowser"
         class="supportedbrowser--alert"
       >
-        Alert - unsupported browser
+        Your browser does not support the Superhero extension.
+        Tipping and commenting will be disabled.
       </div>
       <keep-alive :max="5">
         <router-view :key="$route.fullPath" />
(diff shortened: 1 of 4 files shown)


src/App.vue
```diff
@@ -11,7 +11,8 @@
         v-if="isSupportedBrowser"
         class="supportedbrowser--alert"
       >
-        Alert - unsupported browser
+        Your browser does not support the Superhero extension.
+        Tipping and commenting will be disabled.
```
mradkov, line 15: Can't we just use deeplinks to the webwallet if their browser is not supported?

venimus: It is possible actually, is there an API for that listed?

venimus: It seems however much bigger change for that. Still this is not really related to the PR, as it is for the "supported" ones.

venimus: If we embed the wallet into the website the need of deeplinks would be only for mobile browsers

mradkov: https://github.com/aeternity/superhero-wallet/blob/932e1b6926699174bd1583f5247163ca09bef4f8/tests/pages/index.html

mradkov:
@IvayloPapazov is this finalized?

venimus:
> @venimus is this finalized?

Yes, IMHO

PR #426, opened 2020-04-29
TipsPagination: get lastTipId properly

CedrikNikita:
introduced in #425


mradkov:
Does not seem to be fixing the issue 🤔

CedrikNikita:
> Does not seem to be fixing the issue with getting back from profile page 🤔

Did i connect it to this issue? No.

PR #430, opened 2020-04-29
Added tip user button

stanislav-slavov:
ref #392

The implementation is done without approved design.

### DIFF
--- a/src/components/TipInput.vue
+++ b/src/components/TipInput.vue
@@ -1,6 +1,6 @@
 <template>
   <a
-    v-if="USE_DEEP_LINKS"
+    v-if="USE_DEEP_LINKS && !userAddress"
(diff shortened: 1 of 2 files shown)


src/components/TipInput.vue
```diff
@@ -9,6 +9,16 @@
...
   </a>
+  <a
+    v-else-if="USE_DEEP_LINKS && userAddress"
+    :href="userProfileDeepLink"
+    target="_blank"
+    class="tip__content"
+    @click.stop
+  >
+    <img :src="iconTip">
+    <span class="tip-user-text">Tip User</span>
```
mradkov, line 20:
```suggestion
    <span class="tip-user-text">Tip</span>
```

Issue #428, opened 2020-04-29
Navigation (hamburger) does not appear on smaller widths


venimus:
![image](https://user-images.githubusercontent.com/632282/80584248-0f60b400-8a1a-11ea-98dd-a5de8117d783.png)

I guess the problem comes from not very well defined breakpoints.
In general we should not use "mobile" detection for css breakpoints but rather use mainly generic width (or height in rare cases). So I guess the fix is just to modify the breakpoints of the navigation (loosen the ruleset - remove the mobile specific rules)

PR #431, opened 2020-04-29
Added navigation tablets

stanislav-slavov:
ref #428

There is an issue related to the position of the navigation while resizing(>mobile) on all pages except Tips View.
Feedback would be appreciated.


venimus:
I think we don't need to use `isMobileNavigationHidden` at all, also pixel-aspect-ratio is not very universal.

PR #435, opened 2020-04-30
Refactoring

CedrikNikita:
Should be well tested before merge

### DIFF
--- a/src/components/TipInput.vue
+++ b/src/components/TipInput.vue
@@ -113,7 +113,7 @@ import iconTipped from '../assets/iconTipped.svg';
 import aeternity from '../utils/aeternity';
 import Backend from '../utils/backend';
 import { EventBus } from '../utils/eventBus';
-import util, { USE_DEEP_LINKS } from '../utils/util';
+import util, { USE_DEEP_LINKS, createDeepLinkUrl } from '../utils/util';
 import AeInputAmount from './AeInputAmount.vue';
 import Loading from './Loading.vue';
 import AeButton from './AeButton.vue';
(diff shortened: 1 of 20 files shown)


src/views/TipCommentsView.vue
```diff
@@ -145,12 +145,9 @@ export default {
...
       if (this.USE_DEEP_LINKS) {
-        const url = new URL(`${process.env.VUE_APP_WALLET_URL}/comment`);
-        url.searchParams.set('id', this.tip.id);
-        url.searchParams.set('text', this.newComment);
-        url.searchParams.set('x-success', window.location);
-        url.searchParams.set('x-cancel', window.location);
-        window.location = url;
+        window.location = createDeepLinkUrl(
+          { type: 'comment', id: this.tip.id, text: this.newComment },
+        );
```
venimus, line 150:
```suggestion
        window.location = createDeepLinkUrl(
          { type: 'comment', id: this.tip.id, text: this.newComment, }
        );
```
or better

```suggestion
        window.location = createDeepLinkUrl({
          type: 'comment',
          id: this.tip.id,
          text: this.newComment,
        });
```

src/components/TipInput.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <a
     v-if="USE_DEEP_LINKS && !userAddress"
-    :href="deepLink"
+    :href="createDeepLinkUrl({ type: isRetip ? 'retip' : 'tip', id: tip.id })"
```
venimus, line 4: I would have kept the deepLink function and put the logic there

venimus: Too soon... #439

mradkov:
Can you sync with `develop` and resolve the conflicts? @CedrikNikita , thanks!

src/views/TipCommentsView.vue
venimus: Why? There is no second parameter IMHO this is less right than some formatting
I guess this is some bug in the es-linter when formatting this (weird) way

PR #440, opened 2020-04-30
Extract all locales in file

vmangelovv:


src/views/Privacy.vue
```diff
@@ -6,171 +6,89 @@
...
 
-          <p>
-            I. Introduction
-            Superhero.com LVC, Dr. Grass Str 12, 9490 Vaduz,
-            Liechtenstein (“Superhero.com”, “we”, “us” or “our”)
-            is a company which takes your privacy and the security
-            of your information as a User very seriously.
-          </p>
+          <div class="first">
+            <h5><b>{{ $t('views.Privacy.first') }}</b></h5>
```
venimus, line 16: IMHO this should accept the whole block as HTML (of course json escaped) and put as raw html

PR #442, opened 2020-04-30
Updated texts

mradkov:


src/components/onboarding/steps/Step4.vue
```diff
@@ -19,7 +19,7 @@
     <p class="highlited">
       You can directly support any person's content, cause,<br>
       project that have a public URL with no middleman.<br>
-      (Yes - ANY URL on the web or on <a href="https://superhero.com">www.superhero.com</a>.)
+      (An URL on the web or on <a href="https://superhero.com">www.superhero.com</a>.)
```
venimus, line 22: I think word starting with U should be A URL because of English grammar has that rule if it is pronounced [ju] "U" should not be count as a vowel (because of the "j" actually) in contrast of upload where U is [ʌ]

src/views/CreateProfile.vue
```diff
@@ -148,7 +148,7 @@
             </a>
           </p>
         </expandable-block>
-        <expandable-block :title="'CASHING OUT'">
+        <expandable-block :title="'WITHDRAW'">
```
venimus, line 151: I suggest to use Withdraw to other address or Transfer

PR #459, opened 2020-05-04
Implemented balance update

stanislav-slavov:
ref #451

### DIFF
--- a/src/components/layout/RightSection.vue
+++ b/src/components/layout/RightSection.vue
@@ -30,7 +30,7 @@
           <div v-else>
             <div
               class="balance text-ellipsis"
-              :title="roundAE(balance) + ' AE'"
+              :title="roundAE + ' AE'"
             >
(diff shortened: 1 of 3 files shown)


mradkov:
Its not working well actually. I sent a tip and the balance on the right was not updated.

src/components/layout/RightSection.vue
```diff
@@ -134,8 +134,8 @@ export default {
     selectCurrency(selectedCurrency) {
       this.updateCurrency(selectedCurrency);
     },
-    roundAE(value) {
-      return new BigNumber(value).toFixed(2);
+    roundAE() {
```
venimus, line 137: That would require also to remove the parameter from calls in the template https://github.com/aeternity/superhero-ui/pull/459/files#diff-c2c86662ffc089c65d251de49ee92ae6R33

Issue #475, opened 2020-05-07
Create nested comments functionality


stanislav-slavov:
The scope of this issue is to apply latest designs and functionality for nested comments.
The Tip URL View should display all comments related to a tip, plus their immediate sub-comments with the ability to create new sub-comments.

See the attached screenshot: 
<img width="523" alt="Screenshot 2020-05-07 at 18 54 59" src="https://user-images.githubusercontent.com/7085525/81316506-54c55700-9094-11ea-9cfc-6753e812c92d.png">

PR #476, opened 2020-05-07
Create nested comments functionality

stanislav-slavov:
ref #475


mradkov:
Comments count for tips is not displayed in the main feed. And activity in user profile is emtpy.

PR #482, opened 2020-05-08
replace currency abbr with its symbols, fix user info alignment

vmangelovv:


venimus:
Why not use `Number.toLocaleString('de-DE', { style: 'currency', currency: 'EUR' }` ?

PR #485, opened 2020-05-08
Currency format change

vmangelovv:


src/components/FiatValue.vue
```diff
@@ -3,41 +3,26 @@
...
+      const rate = this.currencyRates.aeternity[this.settings.currency];
       if (!this.amount) {
         return 0;
       }
-      return new BigNumber(this.amount)
-        .multipliedBy(this.currencyRates.aeternity[this.settings.currency]).toFixed(2);
-    },
-    currencySign() {
-      return this.settings && this.settings.currency ? currencySigns[this.settings.currency] : '';
+      return (this.amount * rate).toLocaleString('de-DE', { style: 'currency', currency: this.settings.currency });
```
venimus, line 25: the locale should be from the current language (when implemented) but put it by default for now `en-US` as it matches the one we had

src/components/FiatValue.vue
```diff
@@ -3,41 +3,26 @@
...
+      const rate = this.currencyRates.aeternity[this.settings.currency];
       if (!this.amount) {
         return 0;
       }
-      return new BigNumber(this.amount)
-        .multipliedBy(this.currencyRates.aeternity[this.settings.currency]).toFixed(2);
-    },
-    currencySign() {
-      return this.settings && this.settings.currency ? currencySigns[this.settings.currency] : '';
+      return (this.amount * rate).toLocaleString('en-US', { style: 'currency', currency: this.settings.currency });
```
mradkov, line 25: `en-US` should be based on the selected currency or?

vmangelovv: i think it's based on the selected language which is en by default for now.

Issue #492, opened 2020-05-12
Tip user deeplink


mradkov:
This should be deeplink to the webwallet if no active connection to extension is established.

<img width="545" alt="Screenshot 2020-05-12 at 21 55 43" src="https://user-images.githubusercontent.com/13139371/81733895-5f6c5b80-949b-11ea-8eb7-33a6130d8d71.png">

PR #495, opened 2020-05-13
If a the user is not logged, tip/retip redirect to the wallet

stanislav-slavov:
ref #492

### DIFF
--- a/src/components/TipInput.vue
+++ b/src/components/TipInput.vue
@@ -1,6 +1,6 @@
 <template>
   <a
-    v-if="USE_DEEP_LINKS"
+    v-if="USE_DEEP_LINKS || !canTip"
     :href="deepLink"
     target="_blank"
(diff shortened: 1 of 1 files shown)


src/components/TipInput.vue
```diff
@@ -248,6 +248,9 @@ export default {
       }
       return i18n.t('components.TipInput.totalRetips');
     },
+    canTip() {
+      return this.isLoggedIn && !this.loading.wallet;
```
venimus, line 252: IMHO we need to have a separate case where wallet is loading in this case it should not use deeplink but rather disable the button

stanislav-slavov: There is no indication that a user is being authenticated. The process in not straightforward. I think it would be more confusing if buttons are enabled and disabled on the background. In this implementation are always doing their purpose - tip/retip.

Issue #494, opened 2020-05-13
Tip message to be on top of preview


mradkov:

venimus:
This change also requires to move the bottom tip button near the user profile (which itself requires a redesign)

PR #496, opened 2020-05-13
Set tip message to be ontop of the preview

stanislav-slavov:
ref #494

### DIFF
--- a/src/components/tipRecords/TipComment.vue
+++ b/src/components/tipRecords/TipComment.vue
@@ -248,6 +248,7 @@ export default {
 
     .comment__actions {
       padding-left: 0;
+      padding-top: 0.5rem;
     }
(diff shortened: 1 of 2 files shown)


venimus:
I think a related to this is the other request to move the bottom tip button somewhere else, but we don't have it designed (figured out). Else we have 2 visually the same actions one next to the other and it becomes more confusing. May be do not merge that before it is clear?

mradkov:
> I think a related to this is the other request to move the bottom tip button somewhere else (close to the user image/address), but we don't have it designed (figured out). Else we have 2 visually the same actions one next to the other and it becomes more confusing, which for me will be worse than what we had before. May be do not merge that before it is clear?

Tip button will be removed for now.

PR #514, opened 2020-05-15
Set max length for tip message fields

stanislav-slavov:
ref #504

### DIFF
--- a/src/components/TipInput.vue
+++ b/src/components/TipInput.vue
@@ -77,6 +77,7 @@
           >
             <input
               v-model="message"
+              maxlength="280"
               type="text"
               class="form-control tip__message"
               :placeholder="$t('addMessage')"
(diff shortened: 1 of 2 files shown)


src/components/TipInput.vue
```diff
@@ -77,6 +77,7 @@
           >
             <input
               v-model="message"
+              maxlength="180"
```
mradkov, line 80:
```suggestion
              maxlength="280"
```

src/components/layout/sendTip/SendTip.vue
```diff
@@ -22,6 +22,7 @@
             v-model="sendTipForm.title"
             type="text"
             class="form-control comment"
+            maxlength="180"
```
mradkov, line 25:
```suggestion
            maxlength="280"
```

PR #517, opened 2020-05-15
tip tracing ui

thepiwo:


mradkov:
This is dependant on the backend to be deployed, right?

thepiwo:
Yes it is

PR #528, opened 2020-05-20
Integrate wallet into UI

mradkov:


davidyuk:
@mradkov It works fine with https://github.com/aeternity/superhero-wallet/pull/365

mradkov:
> @mradkov It works fine with [aeternity/superhero-wallet#365](https://github.com/aeternity/superhero-wallet/pull/365)

I will test it out, thanks.

mradkov:
@davidyuk So the integrated wallet should be used for all the actions on the superhero UI side. E.g. My profile (the avatar should be synced - the current logged in account in superhero should be the one from the iframed wallet).

All tx and signature requests should be routed to the iframed wallet.

davidyuk:
@mradkov have you tried it with disabled extension? the extension may conflict with this approach

mradkov:
> @mradkov have you tried it with disabled extension? the extension may conflict with this approach

Trying with removed extension on FF. => Redirects me to the webwallet on a new tab instead of allowing me to confirm signatures from within the integrated wallet.

mradkov:
@davidyuk I think we should deal with https://github.com/aeternity/superhero-ui/issues/606 as the UX really bad when having the extension installed and using the integrated wallet at the same time.

davidyuk:
Until #606 is not clear enough I propose to implement it in this way:
- to wait for extension and to show iframe only if an extension is not available
- If extension available then to behave as before.
I'm testing this approach currently.

src/utils/aeternity.js
```diff
@@ -68,10 +69,12 @@ export const scanForWallets = async () => {
     connectionInfo: { id: 'spy' },
   });
   const detector = await Detector({ connection: scannerConnection });
+  const webWalletTimeout = setTimeout(() => store.commit('useIframeWallet'), 2000);
```
davidyuk, line 72: Timeout is a tricky part, it depends on extension load time. In my case, 2 seconds is enough. I don't think that we can check it in some other way.

.env
```diff
@@ -1 +1 @@
-VUE_APP_WALLET_URL=https://wallet.superhero.com
+VUE_APP_WALLET_URL=https://feature-confirm-connection.wallet.z52da5wt.xyz
```
davidyuk, line 1: We should revert it before making a release

davidyuk:
This implementation depends on https://github.com/aeternity/superhero-wallet/pull/396

src/utils/aeternity.js
mradkov: Lets see how it behaves when testing. Maybe we can think of other way for detection with intervals and fallback after some time 🤔

davidyuk: We can hide iframe if wallet appear later than 2 sec 🙃

mradkov: In terms of UX I'd rather prefer to have some sort of notification (popup) presented to the user so they can confirm wether they want to switch to using the extension. But if we get to agreement for specification and implementation of #606 - then this problem should be obsolete as if you have extension installed should be able to sync it with the integrated wallet. 💭

mradkov:
The integrated wallet does not connect to the voting app 🤔 (based on my testing). @davidyuk can you reproduce?

mradkov:
Updating profile works, but the edit form is not closed after confirm button is clicked.

PR #532, opened 2020-05-20
Added site language switcher

stanislav-slavov:
There are no translations for the languages.
The language is only stored in the vuex.(not bound to backend)
Should appear on mobile.

ref #535


mradkov:
Closing this as its stale.

PR #534, opened 2020-05-20
Added menu for toggling hidden content

stanislav-slavov:
This functionality is a basis for implementing the content language switcher - it will part of the implemented menu.

ref #536


src/locales/zh.json
```diff
@@ -412,7 +412,11 @@
       "SearchPlaceholder": "Search Superhero",
       "SortingLatest": "最新的",
       "SortingHighestRated": "获取♥币最高的 记录排序",
-      "SortingMostPopular": "最受欢迎"
+      "SortingMostPopular": "最受欢迎",
+      "Clear": "Clear",
+      "CloseSearch": "Close Search",
+      "ShowFiltered": "Show filtered content",
+      "HideFiltered": "Hide filtered content"
```
venimus, line 419: I think the best is to use the same text for both states
```suggestion
      "ShowFiltered": "Filter content",
      "HideFiltered": "Filter content"
```
or may be "Safe content only" could also be a good title

src/locales/en.json
```diff
@@ -340,7 +340,9 @@
       "SortingHighestRated": "Highest Rated",
       "SortingMostPopular": "Most Popular",
       "Clear": "Clear",
-      "CloseSearch": "Close Search"
+      "CloseSearch": "Close Search",
+      "ShowFiltered": "Show filtered content",
+      "HideFiltered": "Hide filtered content"
```
venimus, line 345: I think the best is to use the same text for both states
```suggestion
      "ShowFiltered": "Filter content",
      "HideFiltered": "Filter content"
```
or may be "Safe content only" could also be a good title

src/views/TipsList.vue
```diff
@@ -38,12 +38,42 @@
...
+                class="overlay-dots"
+                @click="showMenu = false"
+              />
+              <div class="actions-menu">
+                <div
+                  class="dots"
+                  :class="{ 'show-menu': showMenu }"
+                  @click.stop="showMenu = true"
+                >
+                  •••
```
venimus, line 52: i think this should be a CSS pseudo `:before` with content or img background

src/views/TipsList.vue
```diff
@@ -38,12 +38,42 @@
               >
                 {{ $t('views.TipList.SortingHighestRated') }}
               </a>
+              <div
+                v-if="showMenu"
+                class="overlay-dots"
+                @click="showMenu = false"
+              />
+              <div class="actions-menu">
```
venimus, line 46: This probably needs to be a component with a slot for the dropdown content

src/components/Checkbox.vue
```diff
@@ -0,0 +1,79 @@
...
+  width: 0;
+}
+
+.checkmark {
+  position: absolute;
+  top: 0.15rem;
+  left: 0;
+  height: 0.9rem;
+  width: 0.9rem;
+  background-color: #000;
```
venimus, line 56: this is wrongly put into the design file, the color should be the dark grey used in Trending section for background

src/components/Checkbox.vue
```diff
@@ -0,0 +1,79 @@
...
+    >
+    <span class="checkmark" />
+  </label>
+</template>
+
+<script>
+export default {
+  name: 'Checkbox',
+  props: {
+    getter: { type: Boolean, required: true },
```
venimus, line 16: I guess this should be a `Function`

src/components/Checkbox.vue
```diff
@@ -0,0 +1,79 @@
...
+    text: { type: String, default: '' },
+    setter: { type: Function, required: true },
+  },
+  computed: {
+    model: {
+      get() {
+        return this.getter;
```
venimus, line 23: IMHO this should be `return this.getter()`

stanislav-slavov: get() is a function that returns a value that corresponds to the checkbox state. Probably you don't like the name of the property "this.getter" - I could change it, but the logic is fine

src/views/TipsList.vue
stanislav-slavov: There is an issue for unifying the dropdown styles. In order not to double the work I guess it could be optimized in #549. @mradkov What do you think?

src/views/TipsList.vue
stanislav-slavov: Same here, we should apply this change to a unified place I guess

mradkov:

mradkov:
Probably there should be a check if the tip
was flagged and if the show flagged is on to prevent this from happening. As fallback there could be a redirect in place to the main feed.

src/views/TipsList.vue
mradkov: > There is an issue for unifying the dropdown styles. In order not to double the work I guess it could be optimized in #549. @mradkov What do you think?

Agree

PR #548, opened 2020-05-27
Refactoring

### LINKED ISSUE #601: Avoid unnecessary wrappers in vuex
I have seen two general cases:
- wrapping of mutation with an action that directly calls it, like [here](https://github.com/aeternity/superhero-ui/blob/21cf5575ce07b7dd17d2ec0f596567fa2b8a3932/src/store/actions.js#L18-L20)
- wrapping a state field with a getter that directly returns it, like [here](https://github.com/aeternity/superhero-ui/blob/21cf5575ce07b7dd17d2ec0f596567fa2b8a3932/src/store/getters.js#L3-L5)

I see these problems in this approach:
- easier to make a typo because you have to write more to create mutation/state field
- more difficult to refactor because you have to change the same thing in several places
- the code became less obvious because using getter/action you don't know (have to explicitly check) is it a plain state field/mutation or not 
- more difficult to keep code consistent because it works fine without wrappers (as I understand) and developers may accidentally/intentionally miss wrappers creation

Maybe you have a different opinion on that, it would be nice to discuss it.

CedrikNikita:
Show loading properly in SingleCommentView. Showing loading on top instead of at the bottom of all loaded comments.
image
fixes #601


mradkov:
@CedrikNikita can you resolve the conflicts?

mradkov:
@CedrikNikita can you resolve the conflicts so we can merge this?

PR #555, opened 2020-05-28
Enable HTML5 history mode

davidyuk:
- it should make enabling of SSR easier in future
- consistent with a way is wallet deployed
- URLs without anchor considered as [looking more beautiful](https://router.vuejs.org/guide/essentials/history-mode.html)

### DIFF
--- a/scripts/before-deploy.sh
+++ b/scripts/before-deploy.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -e
 
+cp ./dist/index.html ./dist/404.html
+
 if [[ $CNAME != "" ]]; then
   echo $CNAME > dist/CNAME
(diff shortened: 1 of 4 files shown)


mradkov:
This will affect previously sent tips to URLs in superhero which use the hash mode. How should we deal with that?

davidyuk:
I have added a redirection inspired by [this](https://forum.vuejs.org/t/vue-router-history-mode-with-support-for-legacy-hash-mode-urls/38073/3). So, for example, https://feature-enable-history-mode.aggregator.z52da5wt.xyz/#/tip/1375 should be redirected to https://feature-enable-history-mode.aggregator.z52da5wt.xyz/tip/1375

PR #580, opened 2020-05-29
Integrate league site as iframe

venimus:
temporary will integrate a static site until iframe issue with wordpress is resolved

### DIFF
--- a/src/views/League.vue
+++ b/src/views/League.vue
@@ -1,0 +1,24 @@
+<template>
+  <Page>
+    <div class="league-page">
+      <iframe src="https://aeternity.com/league-iframe/index.html" width="100%" />
+    </div>
+  </Page>
+</template>
+<script>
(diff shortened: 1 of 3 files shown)


src/views/League.vue
```diff
@@ -0,0 +1,24 @@
+<template>
+  <Page>
+    <div class="league-page">
+      <iframe src="https://aeternity.com/league-iframe/index.html" width="100%" />
```
venimus, line 4: this is a mirror copy of the league site, we can replace it once the issue with wordpress blocking iframes is resolved

mradkov:
Some fonts and icons are broken for me in the iframe.

venimus:
> Some fonts and icons are broken for me in the iframe.
> 
> Otherwise looks good to me.

There few more css fixes ongoing on the related repo

PR #588, opened 2020-06-01
Share previews

stanislav-slavov:


package.json
```diff
@@ -7,7 +7,7 @@
     "build:stats": "vue-cli-service build --report-json",
     "test:unit": "vue-cli-service test:unit",
     "test:e2e": "vue-cli-service test:e2e",
-    "lint:vue": "vue-cli-service lint --no-fix",
+    "lint:vue": "vue-cli-service lint --fix",
```
CedrikNikita, line 10:
```suggestion
    "lint:vue": "vue-cli-service lint --no-fix",
```

stanislav-slavov: This is just a testing purpose PR, these change won't be part of it

CedrikNikita: This is marked as a PR not a draft, that's why i have add this comment.

stanislav-slavov: No problem, I haven't marked it as draft because I needed the deployment link after every commit.

PR #592, opened 2020-06-03
Added inline svg support

stanislav-slavov:
ref #589


src/components/layout/Navigation.vue
```diff
@@ -97,43 +107,15 @@ export default {
...
-      background-image: url('../../assets/iconUser.svg');
-    }
-
-    .faq & {
-      background-image: url('../../assets/iconHelp.svg');
+    .faq & path,
+    .profile & path,
+    .tips & path,
+    .league & path {
+      fill: currentColor;
```
davidyuk, line 114: Let's use `currentColor` exactly in svg files

Issue #590, opened 2020-06-03
Can't login with wallet on profile page


davidyuk:
**To Reproduce**
1. Open https://develop.aggregator.z52da5wt.xyz/user-profile/ak_2oBEG11B2GZSWVdVUQy1CYEvgWCfeizEGzjDaQB1YFv4owMcbd
2. Click on 'Login with Web Wallet'
3. Confirm address sharing
4. After redirection back to UI address is not showing on the right bar

PR #593, opened 2020-06-03
Login with web-wallet on every page

stanislav-slavov:
ref #590

### DIFF
--- a/src/components/layout/FooterSection.vue
+++ b/src/components/layout/FooterSection.vue
@@ -52,14 +52,14 @@
     </div>
     <div
       v-if="!isLoggedIn"
-      class="ww-login-footer"
+      class="login-footer"
     >
       <a
         :href="createDeepLinkUrl({ type: 'address' })"
         class="button"
-        :title="$t('components.layout.RightSection.LoginWithWallet')"
+        :title="$t('components.layout.FooterSection.LoginWithWallet')"
       >
-        {{ $t('components.layout.RightSection.LoginWithWallet') }}
+        {{ $t('components.layout.FooterSection.LoginWithWallet') }}
       </a>
     </div>
   </div>
@@ -144,7 +144,7 @@ export default {
       }
     }
 
-    .ww-login-footer {
+    .login-footer {
       display: none;
       line-height: 0.9rem;
       margin-top: 0.4rem;
@@ -156,7 +156,7 @@ export default {
   }
 
   @media (max-width: 1024px) {
-    .footer .ww-login-footer {
+    .footer .login-footer {
       display: block;
     }
   }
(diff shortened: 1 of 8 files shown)


src/components/layout/FooterSection.vue
```diff
@@ -50,13 +50,33 @@
...
+import { mapGetters } from 'vuex';
+import { createDeepLinkUrl } from '../../utils/util';
 
 export default {
   name: 'FooterSection',
+  data() {
+    return {
+      createDeepLinkUrl,
+    };
+  },
```
davidyuk, line 78:
```suggestion
  methods: {
    createDeepLinkUrl,
  },
```

src/components/layout/FooterSection.vue
```diff
@@ -50,13 +50,33 @@
...
+    <div
+      v-if="!isLoggedIn"
+      class="ww-login-footer"
+    >
+      <a
+        :href="createDeepLinkUrl({ type: 'address' })"
+        class="button"
+        :title="$t('components.layout.RightSection.LoginWithWallet')"
+      >
+        {{ $t('components.layout.RightSection.LoginWithWallet') }}
```
davidyuk, line 62: `RightSection` is not related to this component. This is why I think lang keys should be structured by terms semantic (if it is necessary) instead of the component hierarchy.

src/components/layout/FooterSection.vue
```diff
@@ -50,13 +50,33 @@
         <img src="../../assets/iconVenture.svg">
       </a>
     </div>
+    <div
+      v-if="!isLoggedIn"
+      class="ww-login-footer"
```
davidyuk, line 55: What means `ww`? If it is a web wallet, then can't it be used by the Cordova version?

src/components/layout/Page.vue
```diff
@@ -33,6 +36,24 @@ export default {
...
+    };
+  },
+  async created() {
+    if (this.address) {
+      await aeternity.initClient();
+      const balance = await aeternity.client.balance(this.address).catch(() => 0);
+      this.setLoggedInAccount({
+        account: this.address,
+        balance: Util.atomsToAe(balance).toFixed(2),
+      });
```
davidyuk, line 51: Can't it be moved to App.vue as far as there already similar stuff?

src/components/layout/FooterSection.vue
stanislav-slavov: The footer is visualized when the mobile navigation is opened so we could login with web-wallet on mobile. This button is also visualized in the RightSection. I could duplicate the strings for this particular case, thus it would follow the semantic strings approach.

src/components/layout/FooterSection.vue
stanislav-slavov: Yes, this is the mobile version of the button for login with web wallet.

PR #596, opened 2020-06-04
Add version and commit hash to a footer

### LINKED ISSUE #551: Add commit id to footer
Display the release commit id / version in the footer.

CedrikNikita:
fixes #551 
We must come to an agreement on design.

### DIFF
--- a/vue.config.js
+++ b/vue.config.js
@@ -1,6 +1,12 @@
 const path = require('path');
 const FaviconsWebpackPlugin = require('favicons-webpack-plugin');
 
+// eslint-disable-next-line camelcase
+const { npm_package_version } = process.env;
+const commitHash = require('child_process')
+  .execSync('git rev-parse HEAD')
+  .toString().trim();
+
 module.exports = {
   lintOnSave: false,
   css: {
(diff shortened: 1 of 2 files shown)


vue.config.js
```diff
@@ -47,6 +50,24 @@ module.exports = {
...
+      Object.entries(definitions['process.env']).forEach(([k, v]) => {
+        definitions[`process.env.${k}`] = v;
+      });
+      delete definitions['process.env'];
+
+      // eslint-disable-next-line camelcase
+      if (npm_package_version) {
+        definitions['process.env.npm_package_version'] = JSON.stringify(npm_package_version);
+      }
+      const gitRevisionPlugin = new GitRevisionPlugin();
```
davidyuk, line 66: Actually, it is not used as a webpack plugin, maybe we can use something super simple like https://stackoverflow.com/a/35778030/6176994? (but to put require at the top of the file)

vue.config.js
```diff
@@ -47,6 +50,24 @@ module.exports = {
...
     })
     .end()
+    .plugin('define')
+    .tap((options) => {
+      const definitions = { ...options[0] };
+
+      Object.entries(definitions['process.env']).forEach(([k, v]) => {
+        definitions[`process.env.${k}`] = v;
+      });
+      delete definitions['process.env'];
```
davidyuk, line 60: Is this actually used somehow?

mradkov:
> fixes #551
> We must come to an agreement on design.

There is no specific design needed for this. We shoud put the commit hash (probably 6-7 chars of it, with a link to the commit in gh) in the footer. (could be next to the "superhero is opensource" text)

PR #597, opened 2020-06-04
UserProfileView: Call sign-message deeplink

### LINKED ISSUE #550: Edit profile with web wallet
Users should be able to edit their profile via deeplink with the web-wallet.

davidyuk:
depends on https://github.com/aeternity/superhero-wallet/pull/338

Questions that should be resolved to have it working:
- [x] can I log in through deeplink on mobile?
- [x] once logged in through deeplink wallet should remember the address
- [x] how can be detected that the wallet is not present in the desktop version?

fixes #550


mradkov:
> Questions that should be resolved to have it working:
> 
> * [ ]  can I log in through deeplink on mobile?

You should be able to login with deeplink on mobile as well - yes.

> * [ ]  once logged in through deeplink wallet should remember the address

Since there is no active connection to the wallet when login with deeplink - we need to keep the address in the local storage so we allow the user to edit their profile (display the Edit button when they are on the profile page), and display the correct avataar/jdenticon.

> * [ ]  how can be detected that the wallet is not present in the desktop version?

Do you mean the extension or ?

davidyuk:
@mradkov as I got the first two points is not implemented and would be good if somebody helps me

> Do you mean the extension or ?

Yes, right now we have `USE_DEEP_LINKS` and it is always false on desktop. Somehow we should detect that there are no extension and switch it to true then.

stanislav-slavov:
Looks good. There is one issue I found - when we set an empty description we get a "bad request" error.

mradkov:
> Looks good. There is one issue I found - when we set an empty description we get a "bad request" error. I guess we should send empty string value.

@davidyuk  can you have a look?

Also can you resolve the conflicts?

davidyuk:
> There is one issue I found - when we set an empty description we get a "bad request" error. 

I think this issue is on the backend side:
https://github.com/aeternity/tipping-community-backend/pull/133/files

@mradkov I have resolved the conflicts

PR #659, opened 2020-06-22
Use the same screen for superhero and league.aeternity

shapkarin:


src/views/Conference.vue
```diff
@@ -17,15 +18,25 @@ export default {
...
+
+
+    connection.connect(({ room }) => {
+      console.log({ room });
+      this.$router.push({ name: 'conference', params: { room } });
+    });
+
     // eslint-disable-next-line no-new
-    new JitsiMeetExternalAPI('test.league.aeternity.org', {
+    new JitsiMeetExternalAPI('localhost:8080 ', {
```
shapkarin, line 35: what domain should I use? test.league.aeternity.org or league.aeternity.org?

shapkarin:
I will also add a preloader, because now switch from create room page and the current room page is not so smooth.

src/views/Conference.vue
davidyuk: you may extract it to https://github.com/aeternity/superhero-ui/blob/develop/.env and to use league.aeternity.org by default

src/views/Conference.vue
```diff
@@ -17,15 +18,23 @@ export default {
...
   },
-  mounted() {
+  async mounted() {
+    const connection = await BrowserWindowMessageConnection({
+      origin: 'https://localhost:8080',
+    });
+
+    connection.connect(({ room }) => {
+      this.$router.push({ name: 'conference', params: { room } });
+    });
```
davidyuk, line 30: I suggest to don't use this api and do it by ourselves. Actually, it is just `window.addEventListener('message', ...)`

src/router.js
```diff
@@ -103,6 +102,7 @@ const routes = [
     meta: {
       title: 'Superhero League',
     },
+    // children: []
```
davidyuk, line 105: ✂️

src/router.js
```diff
@@ -149,7 +149,7 @@ const routes = [
   {
     path: '/networking',
     name: 'networking',
-    component: Networking,
+    component: Conference,
     meta: {
       title: 'Networking',
     },
```
davidyuk, line 155: Can you combine this route with the next one?

src/views/Conference.vue
shapkarin: @davidyuk true, but I think that to use the same `BrowserWindowMessageConnection` makes code (how to say that)  more consistency.

src/views/Conference.vue
shapkarin: Thanks

src/router.js
shapkarin: thanks

src/views/Conference.vue
shapkarin: Also I didn't import the new dependency for that and use the same at the jitsi side for communion with wallet and superhero-ui . So I think to leave that as it is for the better consistency.

shapkarin: @davidyuk I looked at https://github.com/aeternity/aepp-sdk-js/blob/47179f7a30c39b59110ebec82919ce219367f4b1/es/utils/aepp-wallet-communication/connection/browser-window-message.js and think that you are right. BrowserWindowMessageConnection instance can take more memory.

shapkarin:
we also discuss with @davidyuk about memory leak because I didn't unsubscribe RPC listener

src/router.js
shapkarin: unfortunately aliases does not work with params. But I will try to add `networking` alias to the `conference` maybe

shapkarin: it also doesn't work with `alias: [ '/networking' ],` and `<router-link :to="{ name: 'conference' }">`

src/views/Conference.vue
shapkarin: I will leave `BrowserWindowMessageConnection

src/router.js
davidyuk: My idea was to have a single url with an optional `room` parameter, like:
```
{
    path: '/league/:room?',
    name: 'conference',
    component: Conference,
    props: true,
    meta: {
      title: 'Conference',
    },
  },
```
I didn't mean to use aliases. Can it work in this way?

src/views/Conference.vue
```diff
@@ -1,36 +1,59 @@
...
+
     // eslint-disable-next-line no-new
-    new JitsiMeetExternalAPI('test.league.aeternity.org', {
+    new JitsiMeetExternalAPI(process.env.VUE_APP_JITSI_URL, {
       parentNode: this.$refs.jitsi,
       width: '100%',
       height: '100%',
-      roomName: this.room,
+      roomName: this.room || '',
```
davidyuk, line 45: is this necessary as far as you already enabled `default: ''` above?

src/views/Conference.vue
```diff
@@ -1,36 +1,59 @@
 <template>
   <Page class="league-page-container">
+    <Loading v-show="loading" />
```
davidyuk, line 3: I would suggest using `v-if` by default as far as it works simpler than `v-show`
```suggestion
    <Loading v-if="loading" />
```

src/views/Conference.vue
```diff
@@ -1,36 +1,59 @@
...
+  }),
+  async mounted() {
+    this.connection = await BrowserWindowMessageConnection({
+      origin: `https://${process.env.VUE_APP_JITSI_URL}`,
+    });
+
+    this.connection.connect(({ room }) => {
+      this.$router.push({ name: 'conference', params: { room } });
+    });
+
```
davidyuk, line 39: you may use `$once` to avoid putting `connection` to component's state:
```suggestion

    this.$once('hook:destroyed', () => connection.disconnect());
```

src/views/Conference.vue
shapkarin: I can't use `v-if` because jitsi instance needs some container to make sure that iframe if loaded.

src/views/Conference.vue
shapkarin: Yes, it's the main idea for that realisation.

shapkarin: Thanks

src/router.js
shapkarin: I guess so. Thanks.

src/views/Conference.vue
shapkarin: Thanks. Not sure why but
```js
this.$once('hook:destroyed', () => {
      console.log('destroyed');
    });
```
and also 
```js
destroyed() {
    console.log('destroyed');
  },
```
does not log anything. the same with `beforeDestroy

src/views/Conference.vue
davidyuk: Yes, I got it but seems that the Loader doesn't need `v-show`

src/views/Conference.vue
davidyuk: It is because of KeepAlive around RouterView (https://github.com/aeternity/superhero-ui/issues/664)

src/views/Conference.vue
shapkarin: True

src/views/Conference.vue
shapkarin: Agree that `state management` makes app more predictable.
For this time I will use `deactivated hook` it works.

shapkarin: deactivated hook makes that:
- we unsubscribed 
- next we back to the page
- there is no listener anymore

I will leave it with destroy and it will be subscribed till the keep alive.

shapkarin: Docs says that `max` is `The maximum number of component instances to cache.` not sure when unsubscribed will be called. But at least after we open the page second time it use cache and don't subscribe again.

mradkov:

shapkarin:
@mradkov OK

shapkarin:
@mradkov https://github.com/aeternity/superhero-ui/pull/659#pullrequestreview-435217201

mradkov:
> @mradkov Please check this question [#659 (review)](https://github.com/aeternity/superhero-ui/pull/659#pullrequestreview-435217201)
> It use league.aeternity.org domain. What should I use?

`test.league.aeternity.org` is ok for now.

shapkarin:
@davidyuk 
git reset develop
git add .
git commit -m "..."

Please, can you have a look?

shapkarin:
@davidyuk it looks like I remove some files. I will redo that.

shapkarin:
I guess it's better now

src/router.js
```diff
@@ -182,14 +181,6 @@ const routes = [
...
-  {
-    path: '/networking',
-    name: 'networking',
-    component: Networking,
-    meta: {
-      title: 'Networking',
-    },
-  },
   {
     path: '/league/:room?',
```
shapkarin, line 185: not sure why that is not marked as new now. I got that after `git push --force`

shapkarin:
I found out that deeplinks are not works properly with an iframe. I guess I need to open new tabs to achieve that.

PR #670, opened 2020-06-25
Conference e2e [connectivity tests]

shapkarin:
Also it will be nice to have a test about direct page open. Like with two tabs.


shapkarin:
It's works fine if you define the right `VUE_APP_JITSI_URL`

shapkarin:
depends on #659

cypress.json
```diff
@@ -5,5 +5,6 @@
   "env": {
     "privateKey": "7c6e602a94f30e4ea7edabe4376314f69ba7eaa2f355ecedb339df847b6f0d80575f81ffb0a297b7725dc671da0b1769b1fc5cbe45385c7b5ad1fc2eaf1d609d",
     "publicKey": "ak_fUq2NesPXcYZ1CcqBcGC3StpdnQw3iVxMA3YSeCNAwfN4myQk"
-  }
+  },
+  "chromeWebSecurity": false
```
davidyuk, line 9: is it because of iframes?

tests/e2e/specs/conference.spec.js
```diff
@@ -0,0 +1,28 @@
...
+
+  it('change route by click to the iframe "OK" button', () => {
+    const ROOM = 'SOME_ROOM_NAME';
+
+    getIframeBody().find('#enter_room_field').type(ROOM);
+
+    getIframeBody().find('#enter_room_button').click();
+    cy.window().then(({ location: { pathname } }) => {
+      expect(pathname).to.equal(`/league/${ROOM}`);
+    });
```
davidyuk, line 26:
```suggestion
    cy.location('pathname').should('eq', `/league/${ROOM}`);
```

tests/e2e/specs/conference.spec.js
```diff
@@ -0,0 +1,28 @@
...
+const { expect } = require('chai');
+
+const getIframeDocument = () => cy
+  .get('#jitsiConferenceFrame0')
+  .its('0.contentDocument')
+  .should('exist');
+
+const getIframeBody = () => getIframeDocument()
+  .its('body').should('not.be.undefined')
+  .then(cy.wrap);
```
davidyuk, line 11: For me, it is unclear how it works

cypress.json
mradkov: Jitsi requires to be running in `https` only

davidyuk: Currently, it refers to Jitsi instance at league.aeternity.org, https should work fine there.

shapkarin: it's 
```
To enable cross-domain iframe access, I will set the property chromeWebSecurity to false in the file cypress.json and re-run the test.
```
https://www.cypress.io/blog/2020/02/12/working-with-iframes-in-cypress/

tests/e2e/specs/conference.spec.js
shapkarin: @davidyuk 
I can rewrite as .then(iframeBody => cy.wrap(iframeBody));

shapkarin: maybe you mean about the cy.wrap method itself?

davidyuk: yes, I can't get why do you need that at all, here is a wrap doc for reference: https://docs.cypress.io/api/commands/wrap.html#Syntax

tests/e2e/specs/conference.spec.js
```diff
@@ -0,0 +1,24 @@
...
+    cy.visit('/league');
+  });
+
+  it('change route by click to the iframe "OK" button', () => {
+    const ROOM = 'SOME_ROOM_NAME';
+
+    getIframeBody().find('#enter_room_field').type(ROOM);
+
+    getIframeBody().find('#enter_room_button').click();
+    cy.location('pathname').should('eq', `/league/${ROOM}`);
```
davidyuk, line 22: I propose to use chaining more activally, like:
```suggestion
    cy.get('#jitsiConferenceFrame0')
     .its('0.contentDocument')
     .should('exist')
     .its('body')
     .should('not.be.undefined')
     .find('#enter_room_field')
     .type(ROOM)
     .find('#enter_room_button')
     .click()
     .location('pathname')
     .should('eq', `/league/${ROOM}`)
```

davidyuk: Also, does it test stuff inside of iframe? Isn't it enough to just ensure that iframe has a proper url? (then we will be able to enable chromeWebSecurity back)

mradkov: I think we don't need to test things inside the iframe atm.

When the integrated wallet is merged we can extend the e2e to cover connectivity tests.

shapkarin: @mradkov maybe I can create a separate branch and PR for the test propose `just ensure that iframe has a proper url` and leave this one as `connectivity tests` and future?

shapkarin: @mradkov I started it test `change iframe path by Vue route` in separate branch and will open a new PR. I think it's OK?

src/router.js
```diff
@@ -97,8 +96,8 @@ const routes = [
     },
   },
   {
-    path: '/league',
-    name: 'league',
+    path: '/events',
+    name: 'events',
```
shapkarin, line 100: todo: leave as `league`

PR #684, opened 2020-07-03
Conference e2e [iframe's src attr test]

shapkarin:
test set correct iframe path by Vue route 
e2e without chromeWebSecurity: true

ref to: #670


tests/e2e/specs/conference.spec.js
```diff
@@ -1,24 +1,21 @@
...
-  .its('0.contentDocument')
-  .should('exist');
-
-const getIframeBody = () => getIframeDocument()
-  .its('body').should('not.be.undefined')
-  .then(cy.wrap);
-
 describe('Conference.vue', () => {
+  const ROOM = 'SOME_ROOM_NAME';
+  const isValidName = ROOM.match(/^[^?&:"'%#]+$/);
```
shapkarin, line 3: The name can be renamed. Also added jitsi create room check pattern. Can be used to verify if the name is valid.

mradkov: Yes, this is a good one.

PR #691, opened 2020-07-07
TipsPagination: Combine data fetching, fetch simultaneously

davidyuk:


src/components/TipsPagination.vue
```diff
@@ -119,15 +121,6 @@ export default {
         this.loadingMoreTips = false;
       }
     },
-    async reloadData() {
-      this.loadingTips = true;
-
-      this.tips = await Util.range(1, this.page)
-        .asyncMap(async (page) => Backend
-          .getCacheTips(this.tipSortBy, page, this.address, this.search, this.blacklist));
```
davidyuk, line 127: @thepiwo Was it intentional to make these requests one by one instead of doing them at the same time? I don't think that something can go wrong if the client will request them at the same time, except it has 100 pages opened or so. Also, it can be handled by the backend allowing to accept page range.

shapkarin:
@davidyuk I open an issue https://github.com/aeternity/superhero-ui/issues/693

src/components/TipsPagination.vue
thepiwo: my goal here was to get the order right without much code

src/components/TipsPagination.vue
```diff
@@ -28,12 +28,11 @@
 </template>
 
 <script>
-import { get, isEmpty } from 'lodash-es';
+import { get, isEmpty, times } from 'lodash-es';
```
thepiwo, line 31: I'd prefer not to use lodash at all

src/components/TipsPagination.vue
```diff
@@ -76,10 +75,10 @@ export default {
     this.scroll();
 
     EventBus.$on('reloadData', () => {
-      this.reloadData();
+      this.loadData();
     });
 
-    this.interval = setInterval(() => this.reloadData(), 120 * 1000);
+    this.interval = setInterval(() => this.loadData(), 120 * 1000);
```
thepiwo, line 81: the name `reloadData` was consistent across components

src/components/TipsPagination.vue
thepiwo: But could probably be done using Promise all as well

src/components/TipsPagination.vue
davidyuk: why so? I'm using it in edge cases when js doesn't have something out of the box. Also `lodash-es` uses es6 imports so only used code will appear in the bundle. Or maybe you prefer to use ramda instead? 🙃

thepiwo: I don't have a specific reason, I don't think lodash is needed really. Up to you to decide

PR #692, opened 2020-07-07
TipsPagination: Simplify infinite loading

davidyuk:
I'm not sure that I have tested it enough carefully
SImilar PR: https://github.com/aeternity/aepp-base/pull/1400

### DIFF
--- a/src/components/TipsPagination.vue
+++ b/src/components/TipsPagination.vue
@@ -105,7 +105,7 @@ export default {
       }
     },
     async loadMoreTips() {
-      if (!this.endReached) {
+      if (!this.endReached && !this.loadingMoreTips) {
         this.loadingMoreTips = true;
         const tips = await Backend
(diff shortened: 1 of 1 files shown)


mradkov:
Does not work for me

davidyuk:
I have made detection a less accurate by 100 pixels, I think it should fix the issue. Also, I made it work the same as in [MDN guide](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollHeight#Problems_and_solutions).
For some reason, I don't have overscrolling in Firefox on my MacBook. @mradkov can you check it again?

mradkov:
> I have made detection a less accurate by 100 pixels, I think it should fix the issue. Also, I made it work the same as in [MDN guide](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollHeight#Problems_and_solutions).
> For some reason, I don't have overscrolling in Firefox on my MacBook. @mradkov can you check it again?

Its ok now. :)

PR #694, opened 2020-07-07
Aligned vertically activity buttons on user profile

stanislav-slavov:


src/components/ListOfTipsAndComments.vue
```diff
@@ -9,14 +9,14 @@
...
-          {{ $t('components.ListOfTipsAndComments.Channel') }}
+          <span>{{ $t('components.ListOfTipsAndComments.Channel') }}</span>
         </div>
         <div
           :class="['filter-button', { active: activity === 'activity' }]"
           @click="activity = 'activity'"
         >
           <IconActivity />
-          {{ $t('components.ListOfTipsAndComments.Activity') }}
+          <span>{{ $t('components.ListOfTipsAndComments.Activity') }}</span>
```
davidyuk, line 19: is there any reason why additional `span` needed to enable middle alignment?

stanislav-slavov:
The styles for filter-button would be extracted #696. Closing the PR.

davidyuk:
The vertical alignment of text is a tricky topic and I would like to discuss it. As I know `vertical-align: middle` aligns properly a string like "Xy", but "Xx" would look misaligned and I don't know a way better than defining some constant value for `vertical-align`. And I can't get how an extra `span` can help in this case.

stanislav-slavov:
> The vertical alignment of text is a tricky topic and I would like to discuss it. As I know `vertical-align: middle` aligns properly a string like "Xy", but "Xx" would look misaligned and I don't know a way better than defining some constant value for `vertical-align`. And I can't get how an extra `span` can help in this case.

It helped in this case because the whole content got universal vertical-align: middle. 

The svg already had this style. I added the style and attached it to the span in order to put it to the text. 

https://issue-align-activity-buttons.aggregator.z52da5wt.xyz/user-profile/ak_LAfaJPBcCva1L37wc8ddx17LZtaGNSAQzFJYrq3fkvTSoiqBL

davidyuk:
Ok, I got it. Another approach would be to use `display: flex; align-items: center;` at the root node, but I'm not sure which one is more obvious.

PR #695, opened 2020-07-07
Sticky position has 1px off the top

stanislav-slavov:
The issue occurs on Chrome.

### DIFF
--- a/src/views/TipsList.vue
+++ b/src/views/TipsList.vue
@@ -96,7 +96,6 @@ export default {
 <style lang="scss" scoped>
 .container.wrapper {
   padding-top: 0;
-  margin-top: -0.125rem;
   min-height: 4rem;
 }
 


stanislav-slavov:
> This change looks super unobvious. Is it a Chrome bug?

Yes. It doesn't occur on Firefox.

davidyuk:
This issue should be investigated in detail. Probably it is because of overcomplicated markup.

davidyuk:
@stanislav-slavov for me this issue fixes after removing of this line: https://github.com/aeternity/superhero-ui/blob/cfd50e9545a63a15e205d88b3956b391452028a3/src/views/TipsList.vue#L99
what is the purpose of it?

stanislav-slavov:
> @stanislav-slavov for me this issue fixes after removing of this line:
> 
> https://github.com/aeternity/superhero-ui/blob/cfd50e9545a63a15e205d88b3956b391452028a3/src/views/TipsList.vue#L99
> 
> 
> what is the purpose of it?

The issue still persists and occurs on other pages with the mobile navigation where this style is not present. 
This line seems redundant.

Issue #664, opened 2020-06-25
Avoid wrapping of RouterView with KeepAlive


davidyuk:
As I see in 6b33f1e3ba72e75ac5d5b1e61c89e85e2443a05f, it was made to preserve loaded data and more easily restore scroll position. It makes Vue behave less obviously (https://github.com/aeternity/superhero-ui/pull/659/files#r445270616), I propose to pass displayed data through Vuex, it will allow to cache it and to make components not aware of data fetching and be more stateless.

shapkarin:
Please, have a look at that review https://github.com/aeternity/superhero-ui/pull/659#pullrequestreview-437117148
I'm agree that Vuex will make the whole app more predictable.

PR #698, opened 2020-07-08
Store search query in URL

davidyuk:
blocked by #664 depends on #708


davidyuk:
I have merged it into #708

PR #700, opened 2020-07-08
Added new endpoint for retrieving user comments, sorted desc

stanislav-slavov:


src/components/ListOfTipsAndComments.vue
```diff
@@ -167,13 +167,14 @@ export default {
...
-      Backend.getAllComments()
-        .then((allComments) => {
+      Backend.getUserComments(this.address)
+        .then((userComments) => {
           this.showLoading = false;
           this.error = false;
-          this.comments = allComments.filter(
-            (comment) => comment.author === this.address,
-          );
+          if (userComments.length) {
```
davidyuk, line 174: This condition looks unnecessary. I don't think that it is bad to call `sort` of an empty array. Also, it would be easier to read this without this condition. In general, js is too slow for making such optimizations.

src/components/ListOfTipsAndComments.vue
```diff
@@ -167,13 +167,14 @@ export default {
     },
     reloadData() {
       this.showLoading = true;
-      Backend.getAllComments()
-        .then((allComments) => {
+      Backend.getUserComments(this.address)
+        .then((userComments) => {
```
davidyuk, line 171: I would prefer to use `async`/`await`, but it is out of the scope of this PR.

src/components/ListOfTipsAndComments.vue
```diff
@@ -167,13 +167,14 @@ export default {
...
           this.showLoading = false;
           this.error = false;
-          this.comments = allComments.filter(
-            (comment) => comment.author === this.address,
-          );
+          if (userComments.length) {
+            this.comments = userComments
+              .sort((prev, next) => new Date(next.createdAt) - new Date(prev.createdAt));
```
davidyuk, line 176: `sort` accepts comparator, it should compare two values, but there is no guarantee that the first one will be before the second one. So, I would name them just `a`, `b` 😅

PR #708, opened 2020-07-09
Avoid wrapping of RouterView with KeepAlive

### LINKED ISSUE #664: Avoid wrapping of RouterView with KeepAlive

davidyuk:
closes #664 
depends on #691, #702

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,8 +1,5 @@
 <template>
-  <div
-    id="app"
-    @mousedown="saveScrollPosition"
-  >
+  <div id="app">
     <div
       ref="wrapper"
       class="min-h-screen wrapper"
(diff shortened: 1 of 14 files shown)


davidyuk:
But it should work fine using browser back button and pressing back arrow in the UI, shouldn't it be enough? From my perspective, preserving the scroll position in extra cases will complicate code without significant effort.

PR #709, opened 2020-07-09
Three dots menu works with the web wallet

stanislav-slavov:
To apply all actions coming from the web-wallet in the scope of the tip that is being worked with, I am doing a redirect to the tip page.

### DIFF
--- a/src/components/tipRecords/TipRecord.vue
+++ b/src/components/tipRecords/TipRecord.vue
@@ -45,7 +45,7 @@
               </div>
               <div
                 v-if="account"
-                @click="pinOrUnPinTip(!isTipPinned)"
+                @click="pinOrUnPinTip"
               >
                 {{
                   isTipPinned ?
@@ -151,6 +151,7 @@
 <script>
 import { mapMutations, mapState } from 'vuex';
 import Backend from '../../utils/backend';
+import { createDeepLinkUrl } from '../../utils/util';
 import TipInput from '../TipInput.vue';
 import SuccessModal from '../SuccessModal.vue';
 import FormatDate from './FormatDate.vue';
@@ -183,7 +184,7 @@ export default {
     };
   },
   computed: {
-    ...mapState(['account', 'pinnedItems']),
+    ...mapState(['account', 'pinnedItems', 'useSdkWallet']),
     shouldRender() {
       return !this.tip.url.includes('https://superhero.com/tip/' && '/comment/');
     },
@@ -206,6 +207,15 @@ export default {
       return isPinned !== -1;
     },
   },
+  mounted() {
+    if (!this.$route.path.includes(`/tip/${this.tip.id}`)) {
+      return;
+    }
+    const { method, challenge, signature } = this.$route.query;
+    if (!this.useSdkWallet && method && challenge && signature) {
+      this.applyBackendChanges(method, { challenge, signature });
+    }
+  },
   methods: {
     ...mapMutations(['setPinnedItems']),
     async sendReport() {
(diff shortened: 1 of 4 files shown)


src/components/tipRecords/TipRecord.vue
```diff
@@ -206,45 +210,77 @@ export default {
...
+      } else {
+        const url = new URL(`${window.location.origin}/tip/${this.tip.id}`);
+        url.search = '';
+        window.location = createDeepLinkUrl({
+          type: 'sign-message',
+          message: challenge,
+          'x-success': `${url}?method=${method}&challenge=${challenge}&signature={signature}`,
+        });
+      }
+    },
```
davidyuk, line 263: This code looks similar to one in user profile: https://github.com/aeternity/superhero-ui/blob/37d0eb7ce88f042dd13625345ab2e26301169d9b/src/views/UserProfile.vue#L369-L390
I propose to extract it somehow

src/components/tipRecords/TipRecord.vue
```diff
@@ -206,45 +210,73 @@ export default {
       return isPinned !== -1;
     },
   },
+  mounted() {
+    if (!this.$route.name === 'tip') {
```
davidyuk, line 214: it shouldn't work this way, `this.$route.name` will be negated firstly
```suggestion
    if (this.$route.name !== 'tip') {
```

davidyuk: actually, I'm not sure

davidyuk:
```js
> !('a' === 'b')
true
> !'a' === 'b'
false
```

stanislav-slavov: You're right.

src/utils/applyWalletActions.js
```diff
@@ -0,0 +1,34 @@
...
+      message: challenge,
+      'x-success': `${cbUrl}?method=${method}&challenge=${challenge}&signature={signature}`,
+    });
+  }
+};
+
+export default {
+  applyBackendChanges,
+  backendAuth,
+};
```
davidyuk, line 34: This is not needed as far as they already exported before

src/utils/applyWalletActions.js
```diff
@@ -0,0 +1,34 @@
+import Backend from './backend';
+import { createDeepLinkUrl } from './util';
+
+export const applyBackendChanges = async (method, request, requestParams, cb) => {
+  const args = {
+    sendProfileData: [...requestParams, request],
+    pinItem: [...requestParams, request],
+    unPinItem: [...requestParams, request],
+    sendPostReport: [request],
```
davidyuk, line 9: This method shouldn't be aware of methods specific for components, otherwise, extraction is not so useful, because you have to modify this file to call an additional method in a new component

src/components/tipRecords/TipRecord.vue
```diff
@@ -206,18 +210,21 @@ export default {
...
+  mounted() {
+    if (this.$route.name !== 'tip') {
+      return;
+    }
+    const { method, challenge, signature } = this.$route.query;
+    if (!this.useSdkWallet && method && challenge && signature) {
+      const cb = method === 'sendPostReport' ? this.showSuccessModalCallback : this.updatePinnedItems;
+      const params = method !== 'sendPostReport' ? [this.account] : [];
+      applyBackendChanges(method, { challenge, signature }, params, cb);
+    }
```
davidyuk, line 222: if we have several components with this handler, won't `applyBackendChanges` be called multiple times?

src/components/tipRecords/TipRecord.vue
```diff
@@ -206,18 +210,21 @@ export default {
       return isPinned !== -1;
     },
   },
+  mounted() {
+    if (this.$route.name !== 'tip') {
```
davidyuk, line 214: it disables callback handler, but pin button will be visible on all pages, will it work in case if web wallet used on not 'tip' page? I'm not sure that we need this restriction.

src/components/tipRecords/TipRecord.vue
```diff
@@ -226,25 +233,28 @@ export default {
...
-      }).catch((error) => {
-        console.log(error);
-      });
+    async sendReport() {
+      const postData = {
+        tipId: this.tip.id,
+        author: this.account,
+      };
+      const { challenge } = await Backend.sendPostReport(postData);
+      const url = new URL(`${window.location.origin}/tip/${this.tip.id}`);
```
davidyuk, line 242: isn't it the same as the current url?

src/utils/applyWalletActions.js
```diff
@@ -0,0 +1,34 @@
...
+  }[method];
+  if (!args) throw new Error(`Unknown method: ${method}`);
+  await Backend[method](...args);
+  if (cb) cb();
+};
+
+export const backendAuth = async (method, challenge, client, requestParams, url, cb) => {
```
davidyuk, line 16: we can make these methods as a mixin, to access `sdk` from `this`

src/components/tipRecords/TipRecord.vue
stanislav-slavov: That's why we do a redirect to tip page - to get the tip specific methods. Similar would be for comments.

src/components/tipRecords/TipRecord.vue
stanislav-slavov: Yes, it works, the action redirects to the tip specific page upon wallet redirect. That's why the url(param) for the success callback is passed. Otherwise every tip component on the page would execute the callback.

src/components/tipRecords/TipRecord.vue
stanislav-slavov: > isn't it the same as the current url?

For example you could pin a tip from user-profile. Then the success callback would redirect to tip specific page so only this specific tip methods are executed.

src/components/tipRecords/TipRecord.vue
davidyuk: I mean, the tip page will render several TipRecords and each of them will call `applyBackendChanges`, won't it works in this way?

stanislav-slavov: No, the tip page opens the single tip we would like to pin/report and would execute the pin or report method we are executing. Please check it out.

src/utils/applyWalletActions.js
stanislav-slavov: This would require further refactoring of how client object is initialized and retrieved. 
Here we have two cases:
- with established rpc client connection
- no client connection 
Both are covered so far.
Do you think we can merge this PR and think of restructuring the client initialization?

davidyuk: actually, I have a refactored version, but it depends on #708

mradkov:
Can you resolve the conflicts @stanislav-slavov ?

stanislav-slavov:
Closing the PR. The functionality is covered https://github.com/aeternity/superhero-ui/pull/717

PR #711, opened 2020-07-14
TipRecord: Simplify condition

davidyuk:


davidyuk:
@stanislav-slavov is it computed variable really needed? As I see `SingleComment` doesn't render any `TipRecord`s

stanislav-slavov:
> @stanislav-slavov is it computed variable really needed? As I see `SingleComment` doesn't render any `TipRecord`s

Yeah, it seems redundant now.

PR #717, opened 2020-07-17
Set pinning and reporting to work with web-wallet

### LINKED ISSUE #701: Three dot menu not working with web wallet
Clicking `Report` or `Pin`/`Unpin` not working when logged in with web-wallet.

Screenshot 2020-07-08 at 23 57 09

davidyuk:
Rebased and refactored version of #709 

As I think, this is a more easily reusable version of the conditional calling of web wallet and extension

closes #701

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -25,11 +25,6 @@ import Util, { IS_MOBILE_DEVICE, supportedBrowsers } from './utils/util';
 
 export default {
   name: 'App',
-  data() {
-    return {
-      urlAddress: this.$route.query.address,
-    };
-  },
   computed: {
     ...mapState(['account']),
     isSupportedBrowser() {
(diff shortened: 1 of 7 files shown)


src/utils/backendAuthMixin.js
```diff
@@ -0,0 +1,47 @@
...
+      const {
+        method, address, challenge, signature,
+      } = to.query;
+      if (!method || !address || !challenge || !signature) {
+        next();
+        return;
+      }
+      let succeed = false;
+      try {
+        await Backend[method](address, { challenge, signature });
```
davidyuk, line 19: I'm proposing to postpone rendering until the request is complete to don't refresh data after initial load

src/utils/backendAuthMixin.js
```diff
@@ -0,0 +1,47 @@
...
+import Backend from './backend';
+import { createDeepLinkUrl } from './util';
+import { client } from './aeternity';
+
+export default (ignoreCallInQuery) => ({
+  computed: mapState(['useSdkWallet', 'account']),
+  ...!ignoreCallInQuery && {
+    async beforeRouteEnter(to, from, next) {
+      const {
+        method, address, challenge, signature,
```
davidyuk, line 11: Address can't be taken from the store because `this` is not ready at this moment

src/utils/backendAuthMixin.js
```diff
@@ -0,0 +1,47 @@
...
+        next((vm) => {
+          if (succeed && vm.handleBackendSucceedCall) vm.handleBackendSucceedCall(method);
+          vm.$router.replace(window.location.pathname);
+        });
+      }
+    },
+  },
+  methods: {
+    async backendAuth(method, arg, to) {
+      const { challenge } = await Backend[method](this.account, arg);
```
davidyuk, line 31: I have extracted the first call of the backend to here (it was duplicated before). Also, seems reasonable to update the backend methods signature to always accept the address as a first argument.

src/App.vue
```diff
@@ -94,36 +89,36 @@ export default {
...
     async initialLoad() {
       this.addLoading('initial');
       this.addLoading('wallet');
       await initClient();
       await this.reloadData();
       this.removeLoading('initial');
-      if (this.account) this.removeLoading('wallet');
+      if (this.account) {
+        this.removeLoading('wallet');
+        this.fetchUserData();
```
davidyuk, line 106: Without this data was not fetched if used the web wallet

PR #722, opened 2020-07-21
Hotfix meet bug

shapkarin:
Fix the bug at meet page.
New conference was not inited by the Vue prop change.

### DIFF
--- a/src/views/Conference.vue
+++ b/src/views/Conference.vue
@@ -27,20 +27,15 @@ export default {
   data: () => ({
     loading: true,
   }),
+  watch: {
+    room() {
+      this.loading = true;
+      this.$refs.jitsi.innerHTML = '';
+      this.initJitsi();
+    },
+  },
   async mounted() {
-    // eslint-disable-next-line no-new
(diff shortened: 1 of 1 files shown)


src/views/Conference.vue
```diff
@@ -24,23 +24,24 @@ export default {
   props: {
     room: { type: String, default: '' },
   },
+  watch: {
+    // I don't like this approach
```
venimus, line 28: I don't as well

shapkarin: the problem that by the Vue router change and the prop change we don't trigget mounted method as before.. If I can understand why that happen, I can make some better solution.

shapkarin: @venimus we talked and looks like it's a normal Vue flow - don't trigger mounted at props change. So we have two ways: this one, and the second one is create a separate component that will get that prop and change the router.js.

shapkarin: @venimus maybe there is possible to enter the room by jitsi API and remove that all.

venimus: I was talking about the DOM manipulations actually, but seems you fixed it

PR #734, opened 2020-07-26
Refactor amount components and related store

davidyuk:


src/store/modules/backend.js
```diff
@@ -15,6 +15,10 @@ export default {
     tip: {},
     comment: {},
     stats: null,
+    prices: {},
+  },
+  getters: {
+    minTipAmount: ({ prices: { usd } }) => 0.01 / usd || 0.06,
```
mradkov, line 21: What is this `0.06` for?

The requirement was to be able to tip `0.01` USD or more 🤔

davidyuk: This value will be used until rates are fetched. I have just calculated it by the current rates. I think it should work fine even without a default value, should I remove that? Previously the default value was `0.01` that is also no enough clear.

mradkov: But the prices are already being fetched

davidyuk: I have removed that and rebased in at the top of develop.

src/components/FiatValue.vue
```diff
@@ -16,15 +16,19 @@ export default {
...
   },
-  computed: {
-    ...mapState(['settings', 'currencyRates']),
-    fiatValue() {
-      const rate = this.currencyRates.aeternity[this.settings.currency];
-      if (!this.amount) {
-        return 0;
-      }
-      return (this.amount * rate).toLocaleString('en-US', { style: 'currency', currency: this.settings.currency });
+  computed: mapState({
```
mradkov, line 19: Should we continue using this syntax?

davidyuk: This is a vuex feature. They wouldn't implement that if it has an obviously bad readability or so.

PR #753, opened 2020-08-11
Rename state.account to state.address

davidyuk:
I think this naming is more consistent with other projects. I propose to assume account an object with `address`, `balance`, `name`, and other keys like them.


mradkov:
> I think this naming is more consistent with other projects. I propose to assume account an object with `address`, `balance`, `name`, and other keys like them.

Sounds good. We need to spend more time documenting the code and also creating general documentation for the wallet itself as well.

PR #756, opened 2020-08-14
[Prototype] simultaneous v1, v2 contract use

thepiwo:
This is just supposed to merge to the token tip prototype, iframe wallet should stay disabled for demonstration purposes


thepiwo:
This is just supposed to merge to the token tip prototype, iframe wallet should stay disabled for demonstration purposes

Issue #703, opened 2020-07-09
Fix all exceptions and warnings in developer console after initial load


davidyuk:
![Screenshot 2020-07-09 at 15 12 49](https://user-images.githubusercontent.com/9007851/86999731-a7d39b80-c1f6-11ea-966b-2c2981bad9aa.png)

- [x] fix Avatars package related problems
- [x] fix other problems
except for 404's they can't be avoided

CedrikNikita:
In the current stage of develop, there is no such issues. This issue will be closed, feel free to reopen it if the issue will persist in the future releases.

PR #759, opened 2020-08-17
Fix errors, warning on init load

CedrikNikita:
ref #703

### DIFF
--- a/src/components/ButtonPlain.vue
+++ b/src/components/ButtonPlain.vue
@@ -1,0 +1,41 @@
+<template>
+  <Component
+    :is="!disabled && to ? 'Link' : 'button'"
+    :to="to"
+    :disabled="disabled"
+    class="button-plain"
+    @click="$emit('click', $event)"
+  >
+    <slot />
+  </Component>
+</template>
+
+<script>
+import Link from './Link.vue';
+
+export default {
+  components: { Link },
+  props: {
+    to: { type: [String, Object], default: null },
+    disabled: Boolean,
+  },
+};
+</script>
+
+<style lang="scss" scoped>
+.button-plain {
+  display: inline-block;
+  border: none;
+  background: transparent;
+  padding: 0;
+  text-decoration: none;
+  text-align: center;
+  font: inherit;
+  color: inherit;
+  cursor: pointer;
+
+  &[disabled] {
+    cursor: not-allowed;
+  }
+}
+</style>
--- a/src/components/Link.vue
+++ b/src/components/Link.vue
@@ -1,0 +1,32 @@
+<template>
+  <RouterLink
+    v-if="isLinkOnSameHost"
+    class="link"
+    :to="to"
+  >
+    <slot />
+  </RouterLink>
+  <a
+    v-else
+    class="link"
+    :href="to"
+    target="_self"
+    @click="$emit('click', $event)"
+  >
+    <slot />
+  </a>
+</template>
+
+<script>
+export default {
+  props: {
+    to: { type: [String, Object], required: true },
+  },
+  computed: {
+    isLinkOnSameHost() {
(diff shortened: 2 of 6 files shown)


src/components/MessageInput.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div class="message-input">
-    <Avatar :address="address" />
+    <Avatar :address="address || ''" />
```
davidyuk, line 3: I think it is better to set `''` as a default value of `address` property of Avatar component instead.

Especially, if we assume that 
Screenshot 2020-08-17 at 21 03 46
is a valid avatar of an anonymous user.

src/components/layout/FooterSection.vue
```diff
@@ -61,7 +61,7 @@
       class="login-footer"
     >
       <OutlinedButton
-        :href="addressDeepLink"
+        :href="addressDeepLink.toString()"
```
davidyuk, line 64: Technically, we can make it accept Url instance as well, but that component anyway is not enought flexible now, so I don't know how to handle it better. The best would be to make all buttons to accept full url, router location, click handler (displaying it as <a/button/RouterLink>) as we did in the [Base aepp](https://github.com/aeternity/aepp-base/blob/d17541574773e0b2db3fd1b66b1b0a4ea06ccf4a/src/components/ButtonPlain.vue).

mradkov:
@CedrikNikita can you address the suggested changes by @davidyuk. Thanks

src/components/ButtonPlain.vue
```diff
@@ -0,0 +1,45 @@
...
+  border: none;
+  background: transparent;
+  padding: 0;
+  text-decoration: none;
+  text-align: center;
+  font: inherit;
+  color: inherit;
+  cursor: pointer;
+
+  &:focus:not([data-focus-visible-added]) {
```
davidyuk, line 37: This is for adding an outline when keyboard is used, it depends on https://www.npmjs.com/package/focus-visible package.

src/components/Link.vue
```diff
@@ -0,0 +1,35 @@
...
+  <RouterLink
+    v-if="useRouterLink"
+    class="ae-link"
+    :to="to"
+  >
+    <slot />
+  </RouterLink>
+  <a
+    v-else
+    class="ae-link"
```
davidyuk, line 11:
```suggestion
    class="link"
```
the same above

src/components/Link.vue
```diff
@@ -0,0 +1,35 @@
...
+  props: {
+    to: { type: [String, Object], required: true },
+  },
+  computed: {
+    isLinkOnSameHost() {
+      return typeof this.to === 'object'
+        || (new URL(this.to, window.location)).host === window.location.host;
+    },
+    useRouterLink() {
+      return this.$options.components.RouterLink && this.isLinkOnSameHost;
```
davidyuk, line 31: This was needed to be able to use it without VueRoute, it can be removed in this project

davidyuk (approved):

src/components/Link.vue
```diff
@@ -0,0 +1,35 @@
+<template>
+  <RouterLink
+    v-if="useRouterLink"
```
davidyuk, line 3:
```suggestion
    v-if="isLinkOnSameHost"
```

src/components/Link.vue
```diff
@@ -0,0 +1,35 @@
...
+    class="ae-link"
+    :to="to"
+  >
+    <slot />
+  </RouterLink>
+  <a
+    v-else
+    class="ae-link"
+    :href="to"
+    :target="isLinkOnSameHost ? '_self' : '_blank'"
```
davidyuk, line 13:
```suggestion
    target="_blank"
```

PR #768, opened 2020-08-24
extract constants

thepiwo:


src/utils/aeternity.js
```diff
@@ -33,10 +29,10 @@ export const initClient = async () => {
   try {
     if (window.Cypress) {
       client = await Universal({
-        compilerUrl,
+        compilerUrl: COMPILER_URL,
         nodes: [{
           name: 'testnet',
-          instance: await Node({ url: nodeUrlTestNet, internalUrl: nodeUrlTestNet }),
+          instance: await Node({ url: 'https://testnet.aeternity.io' }),
```
mradkov, line 35: Can this also be extracted?

thepiwo: why would we want to extract a constant for testing?

mradkov: Isn't this also used for mainnet?

mradkov (approved):

stanislav-slavov (approved):

PR #774, opened 2020-08-26
Fix image upload request

### LINKED ISSUE #773: Updating profile image is broken

mradkov:
Issue was the correct headers were not passed when an image is uploaded

fixes #773

### DIFF
--- a/src/utils/backend.js
+++ b/src/utils/backend.js
@@ -1,8 +1,9 @@
 import { BACKEND_URL } from '../config/constants';
 import { wrapTry } from './util';
 
-const backendFetch = (path, ...args) => wrapTry(fetch(`${BACKEND_URL}/${path}`, ...args)
-  .catch((err) => console.error(err)));
+const backendFetch = (path, ...args) => wrapTry(
+  fetch(`${BACKEND_URL}/${path}`, ...args).catch((err) => console.error(err)),
+);
 
 export default class Backend {
   static getTipComments = async (tipId) => backendFetch(`comment/api/tip/${encodeURIComponent(tipId)}`);
@@ -45,8 +46,12 @@ export default class Backend {
 
   static sendProfileData = async (address, postParam) => backendFetch(`profile/${address}`, {
     method: 'post',
-    body: JSON.stringify(postParam),
-    headers: { 'Content-Type': 'application/json' },
+    ...postParam instanceof FormData
+      ? { body: postParam }
+      : {
+        body: JSON.stringify(postParam),
+        headers: { 'Content-Type': 'application/json' },
+      },
   });
 
   static setImage = async (address, data) => {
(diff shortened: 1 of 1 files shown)


src/utils/backend.js
```diff
@@ -43,11 +44,22 @@ export default class Backend {
...
+      body:
+        postParam instanceof FormData ? postParam : JSON.stringify(postParam),
+    };
+
+    Object.assign(
+      request,
+      !(postParam instanceof FormData) && {
+        headers: { 'Content-Type': 'application/json' },
+      },
+    );
```
mradkov, line 59: Here we assign the `headers` prop only in the case where no image was uploaded. 

We can optimize if we find a more clean way to have a null-safe object property assign.

davidyuk: I'm proposing to do it in a way as it was before:
```js
static sendProfileData = async (address, postParam) => backendFetch(`profile/${address}`, {
  method: 'post',
  body: postParam instanceof FormData ? postParam : JSON.stringify(postParam),
  ...!(postParam instanceof FormData) && { headers: { 'Content-Type': 'application/json' } },
});
```

src/utils/backend.js
```diff
@@ -1,8 +1,9 @@
 import { BACKEND_URL } from '../config/constants';
 import { wrapTry } from './util';
 
-const backendFetch = (path, ...args) => wrapTry(fetch(`${BACKEND_URL}/${path}`, ...args)
-  .catch((err) => console.error(err)));
+const backendFetch = (path, ...args) => wrapTry(
+  fetch(`${BACKEND_URL}/${path}`, ...args).catch((err) => console.error(err)),
+);
```
davidyuk, line 6: Would be so nice to remove `console.error(err)` from here, but it is out of the scope of this PR

src/utils/backend.js
```diff
@@ -45,8 +46,8 @@ export default class Backend {
 
   static sendProfileData = async (address, postParam) => backendFetch(`profile/${address}`, {
     method: 'post',
-    body: JSON.stringify(postParam),
-    headers: { 'Content-Type': 'application/json' },
+    body: postParam instanceof FormData ? postParam : JSON.stringify(postParam),
+    ...!(postParam instanceof FormData) && { headers: { 'Content-Type': 'application/json' } },
```
davidyuk, line 50: we can also extract condition 🤩
```suggestion
    ...postParam instanceof FormData
      ? { body: postParam }
      : {
        body: JSON.stringify(postParam),
        headers: { 'Content-Type': 'application/json' }
      },
```
@mradkov can you squash all commits if you decide to apply this one?

mradkov: I will apply it and squash.

thepiwo (approved):

davidyuk (approved):

PR #784, opened 2020-08-31
Maintenance.vue: Remove fixed position to allow scroll on iPhone 6/7/8

### LINKED ISSUE #725: Maintenance page is too big on iPhone 6/7/8 and is not scrollable
Screenshot 2020-07-23 at 13 17 44

etharner:
I propose this solution to fix #725.

### DIFF
--- a/src/views/Maintenance.vue
+++ b/src/views/Maintenance.vue
@@ -25,11 +25,14 @@ export default {
 <style lang="scss" scoped>
 .maintenance {
   background-color: #14151f;
-  position: fixed;
   top: 0;
   bottom: 0;
   left: 0;
   right: 0;
+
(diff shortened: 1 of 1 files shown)


mradkov (approved):
This would work for now, but I think we will need to have more accurate media queries to handle such problems in the future.

src/views/Maintenance.vue
```diff
@@ -25,12 +25,15 @@ export default {
 <style lang="scss" scoped>
 .maintenance {
   background-color: #14151f;
-  position: fixed;
   top: 0;
   bottom: 0;
   left: 0;
   right: 0;
```
davidyuk, line 31: Does it have sense without `fixed`? 🤔

Issue #363, opened 2020-04-21
Adblock completly deactivates aggregator


CedrikNikita:

mradkov:
@CedrikNikita can you confirm this is not happening anymore?

etharner:
Well, I've found that it's related to Russian RuAdList AdBlock filter list, it's blocking "Free DNS services very often used for fraudulent schemes, the spread of intrusive ads and viruses." 😄
![image](https://user-images.githubusercontent.com/7098449/91992436-7c44db80-ed77-11ea-916b-5853f9bb1355.png)
I've disabled all filter lists, add this rule and received `net::ERR_BLOCKED_BY_CLIENT` for all backend (which is on .xyz domain)  endpoints.

davidyuk:
I think we need to move backend to some `.com` subdomain. `z52da5wt.xyz` is too cheap to work properly.

mradkov:
We can use `youdonotneedacapetobeahero.com`?

PR #791, opened 2020-09-02
Fix maintenance errors

etharner:
I propose this fixes for errors hell when you can't fetch data from backend (for example, when you have #363 issue). And I don't fix fetch error, because it's primary reason for other errors.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -51,11 +51,9 @@ export default {
       this.reloadData();
     });
     setInterval(() => this.reloadData(), 120 * 1000);
-    EventBus.$on('backendError', () => {
-      this.$router.push({
-        name: 'maintenance',
-      }).catch((err) => { console.error(err); });
-    });
+    EventBus.$on('backendError', () => this.$route.name !== 'maintenance' && this.$router.push({
+      name: 'maintenance',
+    }).catch((err) => { console.error(err); }));
     await this.initialLoad();
   },
   methods: {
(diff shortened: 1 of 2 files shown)


src/store/modules/backend.js
```diff
@@ -64,7 +64,7 @@ export default {
       commit('tipsReloading', args);
       const value = (await Promise.all(
         times(tipsPageCount[args], (p) => Backend.getCacheTips(p + 1, ...args)),
-      )).flat();
+      )).flat().filter((p) => p);
```
mradkov, line 67: Why do we need `.filter((p) => p)` ?

src/App.vue
```diff
@@ -51,11 +51,9 @@ export default {
       this.reloadData();
     });
     setInterval(() => this.reloadData(), 120 * 1000);
-    EventBus.$on('backendError', () => {
-      this.$router.push({
-        name: 'maintenance',
-      }).catch((err) => { console.error(err); });
-    });
+    EventBus.$on('backendError', () => this.$route.name !== 'maintenance' && this.$router.push({
```
mradkov, line 54: This mitigates the redundant redirect loop, right.

src/store/modules/backend.js
etharner: Because when you don't receive any data from backend `times` function still be called 1 time (because we request 1 page)  and will return `[null]` array, not `null` and check here https://github.com/aeternity/superhero-ui/blob/d335355d1e6ea75b4f226027187c83fca0929ce1/src/components/TipsPagination.vue#L7 will pass and you will have `Cannot get id of null` error

mradkov (approved):

PR #798, opened 2020-09-10
Extended timeout

### LINKED ISSUE #780: SendTip: fail to reload data after tip was send
Stuck like this after tip was send
SendTip
```
util.js:24 Uncaught (in promise) Error: Request is cancelled by timeout
vue.runtime.esm.js:1888 Error: Request is cancelled by timeout
```

AtanasKrondev:
Fixes #780 

Extended requests timeout and after sending tip the reloading is delayed, to ensure enough time the back-end to process it.

### DIFF
--- a/src/components/layout/sendTip/SendTip.vue
+++ b/src/components/layout/sendTip/SendTip.vue
@@ -111,12 +111,12 @@ export default {
         .then(async () => {
           await Backend.cacheInvalidateTips().catch(console.error);
           this.clearTipForm();
-          EventBus.$emit('reloadData');
           this.$store.dispatch('modals/open', {
             name: 'success',
             title: this.$t('components.layout.SendTip.SuccessHeader'),
             body: this.$t('components.layout.SendTip.SuccessText'),
           });
+          setTimeout(() => EventBus.$emit('reloadData'), 5000);
         }).catch((e) => {
           this.sendingTip = false;
           if (e.code && e.code === 4) {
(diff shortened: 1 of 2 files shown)


src/components/layout/sendTip/SendTip.vue
```diff
@@ -111,12 +111,12 @@ export default {
         .then(async () => {
           await Backend.cacheInvalidateTips().catch(console.error);
           this.clearTipForm();
-          EventBus.$emit('reloadData');
           this.$store.dispatch('modals/open', {
             name: 'success',
             title: this.$t('components.layout.SendTip.SuccessHeader'),
             body: this.$t('components.layout.SendTip.SuccessText'),
           });
+          setTimeout(() => EventBus.$emit('reloadData'), 5000);
```
mradkov, line 119: Why is this timeout needed here?

AtanasKrondev: The backend itself take about 5 secs to process the tip, so it was suggested not to reload immediately after sending the tip, but delay a little bit.

mradkov: Is delay of `5000` ms enough?

AtanasKrondev: with 5000 ms delay for sending the reload request plus 12 000 ms timeout before throwing the error even for slow connections should be fine.

mradkov (approved):

Issue #767, opened 2020-08-20
Tips not rendered correctly


mradkov:
![8cf8b7ba-a3cb-443d-899d-bc655aa901c0](https://user-images.githubusercontent.com/13139371/90773690-f64c7d80-e2fe-11ea-8d5e-5572a1f2f77d.png)

Could be because of lack of preview returned by the backend? cc @kenodressel

mradkov:
https://superhero.com/tip/1806
 https://superhero.com/tip/1803
 https://superhero.com/tip/1804

davidyuk:
Is it build according to designs? As I see, the amount of tips should be displayed also outside of the preview area:

![Screenshot 2020-08-24 at 16 03 21](https://user-images.githubusercontent.com/9007851/91009167-56a32e00-e623-11ea-897d-36494a2d5bb2.png)

Also, there is a question about it: should the amount of tips in preview be always the same as outside of preview?

mradkov:
> Is it build according to designs? As I see, the amount of tips should be displayed also outside of the preview area:
> 
> 
> 
> ![Screenshot 2020-08-24 at 16 03 21](https://user-images.githubusercontent.com/9007851/91009167-56a32e00-e623-11ea-897d-36494a2d5bb2.png)
> 
> 
> 
> Also, there is a question about it: should the amount of tips in preview be always the same as outside of preview?

The design you are looking at is outdated. 

Only one tip amount is to be present now - the one inside the preview box. It represents the aggregated sum of tips value to that URL.

etharner:
@mradkov Is it still a viable issue?

mradkov:
This should be closed. It will be handled on the backend side by https://github.com/aeternity/tipping-community-backend/pull/172

Issue #765, opened 2020-08-18
Tip amount not visible if no preview is fetched


kenodressel:
If there is no preview available (for new tips or due to backend issues) the tip amount is not displayed.
![image](https://user-images.githubusercontent.com/765921/90494472-d3299e80-e143-11ea-9614-21b720079a88.png)

AtanasKrondev:
Is it still viable issue?

mradkov:
I think this will be automatically resolved with https://github.com/aeternity/tipping-community-backend/pull/172. Correct me if I'm wrong @kenodressel

kenodressel:
Happy to correct you :) It will be less of a problem but there is still a number of tips we are unable to parse so the fallback is still relevant.

PR #799, opened 2020-09-11
Add default protocol for the tip url

CedrikNikita:
ref #765,
ref #767

### DIFF
--- a/src/config/constants.js
+++ b/src/config/constants.js
@@ -2,6 +2,8 @@ export const BACKEND_URL = window.Cypress
   ? 'https://testnet.superhero.aeternity.art'
   : 'https://raendom-backend.z52da5wt.xyz';
 
+export const PROTOCOL_DEFAULT = 'https:';
+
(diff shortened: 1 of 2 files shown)


mradkov:
I like the suggestion, but there are a lot of pages on the web still not using SSL (also a lot of tips in superhero are such).

Let's support both `http://` and `https://` for now, and consider everything else invalid.

When we extend the tipping protocol to support other links such as magnet we can extend the list of supported protocols.

src/config/constants.js
```diff
@@ -2,6 +2,8 @@ export const BACKEND_URL = window.Cypress
   ? 'https://testnet.superhero.aeternity.art'
   : 'https://raendom-backend.z52da5wt.xyz';
 
+export const PROTOCOL_DEFAULT = 'https:';
```
davidyuk, line 5: I'm proposing just to use `http` be default

thepiwo:
whats the state of this?

PR #803, opened 2020-09-14
Tip amount display for tips with no preview generated

### LINKED ISSUE #765: Tip amount not visible if no preview is fetched

AtanasKrondev:
Fixes #765 
Displaying tip amount even if no preview is available


mradkov (approved):
Looks like a good quick fix.

PR #806, opened 2020-09-17
Tips rich preview

AtanasKrondev:
If the tipped URL is an YouTube video rich preview with embedded video is presented.

Related to: https://forum.aeternity.com/t/feature-rich-content-previews/6977/10

Example:

YouTube:

Screenshot 2020-09-17 at 10 29 53

https://feature-rich-content-preview.aggregator.z52da5wt.xyz/tip/1483

SoundCloud:

Twitter:

Others:

No preview:

User profile preview on avatar hover:

_Split to #814_


mradkov:
At first sight, this looks very good! We need to think of a way to personalize the user preferences of embedding the rich content as some of those are touching very much user privacy via cookies, which by nature is against the product's vision. 

Maybe we need to have this turned off by default so when user wants to have embedded videos or other rich preview content - they need to explicitly agree with this.

stanislav-slavov (changes requested):
The code looks good. 
I saw a visual issue 
Screenshot 2020-09-17 at 16 37 37
The preview is overlapping the sticky menu on top.
And here are some suggestions:
- The code could be splitted into multiple contextual commits.
- I think we should disable suggestions for other videos when one is finished.

AtanasKrondev:
The z-index of the video should be fixed now. Unfortunately it's not possible to get rid of YouTube related videos. `&rel=0` doesn't completely remove them, but at least shows only videos from the same channel.

mradkov (changes requested):
There is a wide black space on lower resolutions, can you check and fix it?

Screenshot 2020-09-18 at 9 51 42
Screenshot 2020-09-18 at 9 51 36

AtanasKrondev:
The youtube generated thumbnails have the black strips like this
hqdefault

If we lower the height of the preview card on mobile then there will be not much space for the description.

AtanasKrondev:
image

What do you guys think?

Without the stripes and slightly overlapping the description box, but readable. According to the design the first row is always `YOUTUBE.COM` so there won't be a case the button covering any text.

mradkov:
> image
> 
> What do you guys think?
> 
> Without the stripes and slightly overlapping the description box, but readable. According to the design the first row is always `YOUTUBE.COM` so there won't be a case the button covering any text.

Looks good to me.

stanislav-slavov (approved):

CedrikNikita (changes requested):
Most of the styles and half of the structure of TipRichPreview is copied froom TipRecord component, can we somehow avoid copying the same styles and structure. We can extract styles in the separate files, if needed, or include the component structure into the TipRecord itself.

src/assets/buttonPlay.svg
```diff
@@ -0,0 +1,36 @@
...
+<g>
+</g>
+<g>
+</g>
+<g>
+</g>
+<g>
+</g>
+<g>
+</g>
```
mradkov, line 35:
```suggestion
```

Let's clean up this SVG 🤔

mradkov:
There is some delay in displaying the User profile on hover, can we remove it. Also, it will be nice to create a separate PR for this as it is a separate feature. (looks good though)
ezgif-1-996b6928c2a1

AtanasKrondev:
> There is some delay in displaying the User profile on hover and also it is vanishing with delay on mouse out. Can we remove it (the delay)? Also, it will be nice to create a separate PR for this as it is a separate feature. (looks good though)

Split to #814 with reduced animation time and delay.

src/components/tipRecords/SoundCloudPlayer.vue
```diff
@@ -0,0 +1,224 @@
...
+    playUrl() {
+      return `${this.scApiUrl}?url=${this.tip.url}`;
+    },
+    displayProgress() {
+      return moment.utc(this.progress).format(this.progress < 3600000 ? 'mm:ss' : 'h:mm:ss');
+    },
+    displayDuration() {
+      return moment.utc(this.duration).format(this.duration < 3600000 ? 'mm:ss' : 'h:mm:ss');
+    },
+  },
```
davidyuk, line 92: `moment` the way how it is imported consumes significant part of the bundle:
https://feature-rich-content-preview.aggregator.z52da5wt.xyz/report.html
https://develop.aggregator.z52da5wt.xyz/report.html

I propose to implement this function by ourselves or to use a Date object:
```suggestion
  },
  filters: {
    formatTime: timestamp => {
      const seconds = Math.floor(timestamp / 1000);
      const minutes = Math.floor(seconds / 60);
      const hours = Math.floor(minutes / 60);
      return [
        ...hours ? [hours] : [],
        ...[minutes % 60, seconds % 60].map(t => t.toString().padStart(2, '0')),
      ].join(':');
    },
    // or using a Date object:    
    formatTime: timestamp => new Date(timestamp).toLocaleTimeString([], {
      ...timestamp > 60 * 60 * 1000  ? { hour: 'numeric' },
      minute: '2-digit',
      second: '2-digit',
      timeZone: 'UTC',
    }),
  },
```

Both `Date` and `moment` are not handling the big values properly. For example,
```
> moment.utc(100000000).format('h:mm:ss')
3:46:40 // instead of 27:46:40
```
I'm not sure is it the case of SoundCloud, so I'm not sure which implementation is better.

davidyuk (commented):
This is only moment-related remarks, I will finish the review later.

src/components/tipRecords/TipPreview.vue
```diff
@@ -0,0 +1,526 @@
...
+        .filter((r) => r.test(tip.url)).length > 0;
+    },
+    isPreviewToBeVisualized(tip) {
+      return typeof tip !== 'undefined' && tip !== null
+        && typeof tip.preview !== 'undefined'
+        && (
+          (tip.preview.description !== null && tip.preview.description.length > 0)
+          || (tip.preview.title !== null && tip.preview.title.length > 0)
+        );
+    },
```
davidyuk, line 214: I would make it a computed variable:
```suggestion
    isPreviewToBeVisualized() {
      return this.tip.preview && (this.tip.preview.description || this.tip.preview.title);
    },
```

src/components/tipRecords/SoundCloudEmbed.vue
```diff
@@ -0,0 +1,110 @@
...
+    },
+    loadScript(url) {
+      const scripts = [].slice.call(document.getElementsByTagName('script')).filter((script) => script.src === url);
+      if (scripts.length) {
+        return;
+      }
+      const script = document.createElement('script');
+      script.type = 'text/javascript';
+      script.src = url;
+      document.getElementsByTagName('body')[0].appendChild(script);
```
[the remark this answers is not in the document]
AtanasKrondev: I agree that it's a bad approach.

SoundCloud has HTTP API and there are some unofficial npm packages but you need to have "CLIENT_ID" for them. Currently SoundCloud do not allow registration of new IDs and none of the IDs I found online seems to work anymore.

https://soundcloud.com/you/apps/new

src/components/tipRecords/SoundCloudPlayer.vue
```diff
@@ -0,0 +1,224 @@
...
+        v-else-if="!isPlaying"
+        class="play-button"
+        src="../../assets/buttonPlay.svg"
+        @click="togglePlay"
+      >
+    </div>
+    <div
+      ref="soundWave"
+      class="sound-wave"
+      @click="mousePosition"
```
davidyuk, line 34: the handler name is unclear

src/components/tipRecords/TipPreview.vue
```diff
@@ -0,0 +1,526 @@
...
+      max-width: 100%;
+      width: 100%;
+      position: absolute;
+      right: 0;
+      top: 0;
+      border: none;
+      height: auto;
+      cursor: pointer;
+      -webkit-transition: 0.4s all;
+      -moz-transition: 0.4s all;
```
davidyuk, line 414: Don't we have autoprefixer working in this repository?🤔

src/components/tipRecords/SoundCloudPlayer.vue
```diff
@@ -0,0 +1,224 @@
...
+        />
+      </div>
+      <div class="wave-wrapper flipped">
+        <img
+          class="wave-grey"
+          src="../../assets/soundWave.svg"
+        >
+        <div
+          class="wave-green"
+          :style="{ 'width': `${position}%`, 'background-size': `${1 / position * 10000}% 100%` }"
```
davidyuk, line 53: Let's extract it to computed variables.

src/components/tipRecords/SoundCloudEmbed.vue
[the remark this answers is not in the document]
davidyuk: Actually, there is an official package: https://www.npmjs.com/package/soundcloud (according to https://developers.soundcloud.com/docs/api/sdks), but anyway it won't work without api token.

But, the iframe widget doesn't require the api token. I have found [a wrapper](https://www.npmjs.com/package/soundcloud-widget) around https://w.soundcloud.com/player/api.js, have you tried it?

src/components/tipRecords/TipPreview.vue
```diff
@@ -0,0 +1,526 @@
...
+    SoundCloudEmbed,
+  },
+  props: {
+    tip: { type: Object, required: true },
+    goToTip: { type: Function, default: () => {} },
+    tipUrl: { type: String, default: '' },
+  },
+  data() {
+    return {
+      noPreview: 'No preview available for this URL',
```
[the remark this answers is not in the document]
AtanasKrondev: eslint is complaining about raw text in the template and later on should be replaced with the localization for the different languages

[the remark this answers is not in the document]
davidyuk: So, use eslint-disable-next-line. Your approach is more confusing.

davidyuk:
Doesn't look like that all my remarks are resolved, is it ready for review? @AtanasKrondev

AtanasKrondev:
It should be now @davidyuk

mradkov (approved):
I think this can be merged. The consent page will be implemented in another PR.

PR #809, opened 2020-09-21
Use vuex instead EventBus

### LINKED ISSUE #660: Use vuex instead EventBus pattern
Event bus was introduced in https://github.com/aeternity/superhero-ui/commit/18eafe33fd1bacccadcda9ba6b61440c570f6233

IvayloPapazov:
close #660

### DIFF
--- a/src/components/ListOfTipsAndComments.vue
+++ b/src/components/ListOfTipsAndComments.vue
@@ -92,7 +92,6 @@
 <script>
 import { mapState } from 'vuex';
 import Backend from '../utils/backend';
-import { EventBus } from '../utils/eventBus';
 import Loading from './Loading.vue';
 import TipsPagination from './TipsPagination.vue';
 import TipComment from './tipRecords/TipComment.vue';
@@ -120,7 +119,7 @@ export default {
     userPinnedItems: [],
   }),
   computed: {
-    ...mapState(['loading']),
+    ...mapState(['loading', 'reloading']),
     ...mapState({ currentAddress: 'address' }),
     pinnedItems() {
       return this.address === this.currentAddress
@@ -137,8 +136,11 @@ export default {
   mounted() {
     this.reloadData();
 
-    EventBus.$on('reloadData', () => {
-      this.reloadData();
+    this.$watch(() => this.reloading, (reload) => {
+      if (reload) {
+        this.reloadData();
+        this.$store.commit('reloading', false);
+      }
     });
 
     if (this.address !== this.currentAddress) {
--- a/src/components/SendComment.vue
+++ b/src/components/SendComment.vue
@@ -15,7 +15,6 @@
 </template>
 
 <script>
-import { EventBus } from '../utils/eventBus';
 import { createDeepLinkUrl } from '../utils/util';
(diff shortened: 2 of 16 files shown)


mradkov:
Some tests are failing, can you address that?

src/components/ListOfTipsAndComments.vue
```diff
@@ -137,8 +136,11 @@ export default {
   mounted() {
     this.reloadData();
 
-    EventBus.$on('reloadData', () => {
-      this.reloadData();
+    this.$watch(() => this.reloading, (reload) => {
```
thepiwo, line 139: looks like overhead to me, whats the benefit of this change?

src/components/SendComment.vue
```diff
@@ -53,7 +52,7 @@ export default {
         (data) => client.signMessage(data),
         this.parentId,
       );
-      EventBus.$emit('reloadData');
+      this.$store.commit('reloading', true);
```
thepiwo, line 55: with committing to the store to me its not implicit whats happening, emitting some event is way more clear

src/store/index.js
```diff
@@ -5,6 +5,7 @@ import mutations from './mutations';
 import backend from './modules/backend';
 import persistState from './plugins/persistState';
 import modals from './plugins/modals';
+/* eslint import/no-cycle: [2, { maxDepth: 1 }] */
```
thepiwo, line 8: we should avoid cyclic dependencies

src/components/ListOfTipsAndComments.vue
IvayloPapazov: If I understood correctly the removal of the event bus.

src/store/index.js
IvayloPapazov: I agree. To avoid this, most of the current implementation needs to be changed.

src/components/SendComment.vue
IvayloPapazov: What is your proposal? Should we leave the old implementation?

CedrikNikita:
As i undestood, the main idea of dropping EvenBus from Denis was to drop this workaround behaviour. 
Denis words:
`I think the store should itself update the state (of the component that is displayed) after calling the action.`

thepiwo:
As @CedrikNikita @davidyuk say, I think we should keep a event emit if we need event emit, but have the store update the state

src/components/ListOfTipsAndComments.vue
thepiwo: the benefit of removing the event bus is removing the event bus, whats the benefit of this change actually?

src/components/SendComment.vue
thepiwo: my proposal is to remove the event bus, but not by emulating an event bus with vuex, but use vuex as intended or imagined (cc @davidyuk )

src/store/index.js
thepiwo: yes, that would be necessary

davidyuk:
will be reimplemented in #817, #819, and future PRs

PR #814, opened 2020-09-29
User preview on avatar hover

AtanasKrondev:
_Split from #806_ 

User profile preview on avatar hover:


mradkov:
ezgif-7-917ce98765b8

I don't like how it is revealed on stages. Is it because of the backend request taking more time or sth else?

AtanasKrondev:
> I don't like how it is revealed on stages. Is it because of the backend request taking more time or sth else?

Yes, it makes additional request and some calculations for the stats - it's the same component as in user profile page.
A spinner may be added until all the information is retrieved... or skip entirely the second part with the stats from this preview...

mradkov:
> > I don't like how it is revealed on stages. Is it because of the backend request taking more time or sth else?
> 
> Yes, it makes additional request and some calculations for the stats - it's the same component as in user profile page.
> A spinner may be added until all the information is retrieved... or skip entirely the second part with the stats from this preview...

Let's bring this up for discussion with the design team. cc @onvisions

onvisions:
Ok. It's not ok to load in 2 steps.

Actually I suggested in a [forum post](https://forum.aeternity.com/t/feature-rich-content-previews/6977/23) before reading your question here that we add some delay between the mouse-hover event and the ID card display. This is how Twitter does in order not to trigger by accident displaying of user ID cards while the user is trying to interact with other UI elements. May be this delay of 2-3 seconds will also solve the issue mentioned above?

Another part of this forum suggestion is to use the compact version (intended for mobile).

mradkov:
> Ok. It's not ok to load in 2 steps.
> 
> 
> 
> Actually I suggested in a [forum post](https://forum.aeternity.com/t/feature-rich-content-previews/6977/23) before reading your question here that we add some delay between the mouse-hover event and the ID card display. This is how Twitter does in order not to trigger by accident displaying of user ID cards while the user is trying to interact with other UI elements. May be this delay of 2-3 seconds will also solve the issue mentioned above? 
> 
> 
> 
> Another part of this forum suggestion is to use the compact version (intended for mobile). 
> 
> 
> 
> **Test it on [Twitter](https://twitter.com)!**
> 
> 
> 
> **Screenshot:**
> 
> twitter
> 
> 
> 
> 
> 
> 
> 
> 

Isn't 2-3sec too much?

src/components/UserInfo.vue
```diff
@@ -220,7 +219,15 @@ export default {
   data() {
     return {
       maxLength: 250,
-      userStats: null,
+      userStats: {
```
AtanasKrondev, line 222: Default values for user stats ensures the whole component will be visualized even if the data is delayed.

onvisions:
@mradkov @AtanasKrondev 
2-3 sec. is really too much. Even now if it is 1.5 sec. it seems to me it is too much. But it looks cool though! :)
Another suggestion (easy for implementation) is that we have the **bigger user avatar within the ID card linking to User Profile** page.
At some point all this labels showing "Tips Sent" , "Tips Claimed" will be **links leading to User Profile/ Activity / respective filter loaded** ... (we should add more interactive options to the users not only displaying static info in the ID card).

I found a **problem (bug?) with auto-hiding** the ID card and I managed to display 2 at the same time. 
2atatime

src/components/UserHoverCard.vue
```diff
@@ -0,0 +1,101 @@
...
+  <Transition
+    v-if="hover"
+    name="fade"
+  >
+    <div
+      v-show="show && isActive"
+      class="modal-container"
+      @click.stop
+      @mouseover="mousestay"
+      @mouseleave="mouseleave"
```
davidyuk, line 11:
```suggestion
      @mouseover="$emit('mouseover', $event)"
      @mouseleave="$emit('mouseleave', $event)"
```

src/components/UserHoverCard.vue
```diff
@@ -0,0 +1,101 @@
+<template>
+  <Transition
+    v-if="hover"
+    name="fade"
+  >
+    <div
+      v-show="show && isActive"
+      class="modal-container"
```
davidyuk, line 8:
```suggestion
      class="user-hover-card"
```

src/components/tipRecords/AuthorAndDate.vue
```diff
@@ -1,48 +1,101 @@
...
   </div>
 </template>
 
 <script>
+import { mapMutations, mapState } from 'vuex';
 import FormatDate from './FormatDate.vue';
 import Avatar from '../Avatar.vue';
+import UserHoverCard from '../UserHoverCard.vue';
+
+let timeout;
```
davidyuk, line 55: I don't think that we need the same `timeout` variable for all components. Consider moving it into `data`.

src/components/UserInfo.vue
```diff
@@ -0,0 +1,771 @@
+<template>
+  <div class="profile__page">
```
davidyuk, line 2:
```suggestion
  <div class="user-info">
```
🤷‍♀️

src/components/UserInfo.vue
```diff
@@ -0,0 +1,771 @@
...
+import TipInput from './TipInput.vue';
+import SuccessIcon from '../assets/verifiedUrl.svg';
+
+export default {
+  components: {
+    AeAmountFiat,
+    Avatar,
+    TipInput,
+  },
+  mixins: [backendAuthMixin()],
```
davidyuk, line 215: I would leave it in a page component.

src/components/tipRecords/AuthorAndDate.vue
```diff
@@ -1,48 +1,101 @@
...
+      return this.currentUserInfo === this.showUserInfo;
+    },
+  },
+  methods: {
+    ...mapMutations(['setShowUserInfo']),
+    mouseover() {
+      this.setShowUserInfo(this.currentUserInfo);
+      this.hover = true;
+      clearTimeout(timeout);
+      timeout = setTimeout(() => { this.show = true; }, 500);
```
davidyuk, line 89: can we use `transition-delay` instead?

src/components/tipRecords/AuthorAndDate.vue
```diff
@@ -1,48 +1,101 @@
...
+      </span>
+    </div>
   </div>
 </template>
 
 <script>
+import { mapMutations, mapState } from 'vuex';
 import FormatDate from './FormatDate.vue';
 import Avatar from '../Avatar.vue';
+import UserHoverCard from '../UserHoverCard.vue';
```
davidyuk, line 53: I would rename this component to UserCard

src/components/UserHoverCard.vue
```diff
@@ -0,0 +1,101 @@
...
+  position: absolute;
+  z-index: 99;
+  overflow: hidden;
+  cursor: default;
+  box-shadow: 4px 4px 4px 0 rgba(0, 0, 0, 0.3);
+  background-color: $thumbnail_background_color;
+}
+
+.fade-enter-active,
+.fade-leave-active {
```
davidyuk, line 52: I'm suggesting to use scss nesting more actively:
```suggestion

  &.fade-enter-active,
  &.fade-leave-active {
```

src/components/tipRecords/AuthorAndDate.vue
```diff
@@ -1,48 +1,101 @@
...
+  <div class="wrapper">
+    <UserHoverCard
+      :address="address"
+      :hover="hover"
+      :show="show"
+      :is-active="isActive"
+      :mouseover="mouseover"
+      :mouseleave="mouseleave"
+      :mousestay="mousestay"
+    />
```
davidyuk, line 11: Moving transition out from UserHoverCard, and using `transition-delay` it should be possible to simplify it to:
```suggestion
    <Transition ...>
      <UserCard
        v-if="hover"
        :address="address"
        @mouseover="mouseover"
        @mouseleave="mouseleave"
      />
    </Transition>
```

davidyuk (commented):
Actually, user card designs look differently, are you sure that it should share the layout of UserProfile component?

src/components/tipRecords/AuthorAndDate.vue
```diff
@@ -1,48 +1,101 @@
...
+    return {
+      hover: false,
+      show: false,
+    };
+  },
+  computed: {
+    ...mapState(['showUserInfo']),
+    currentUserInfo() {
+      return this.address + this.parrent;
+    },
```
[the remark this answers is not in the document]
AtanasKrondev: This is to ensure only one card is presented at a time as suggested above.

src/components/tipRecords/AuthorAndDate.vue
```diff
@@ -1,48 +1,101 @@
...
+  methods: {
+    ...mapMutations(['setShowUserInfo']),
+    mouseover() {
+      this.setShowUserInfo(this.currentUserInfo);
+      this.hover = true;
+      clearTimeout(timeout);
+      timeout = setTimeout(() => { this.show = true; }, 500);
+    },
+    mouseleave() {
+      timeout = setTimeout(() => { this.hover = false; this.show = false; }, 500);
```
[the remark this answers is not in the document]
AtanasKrondev: For UX reasons, if there isn't any delay it will be annoying for the user the information to disappear on any unintended mouse movement.

src/components/tipRecords/AuthorAndDate.vue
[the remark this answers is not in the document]
davidyuk: As I got, the timeout was bigger before. But now, will be noticeable/possible to open two cards if the first will be hidden in 0.5s?

src/components/tipRecords/AuthorAndDate.vue
```diff
@@ -1,54 +1,96 @@
...
   },
   props: {
     address: { type: String, required: true },
     name: { type: String, default: '' },
+    parrent: { type: String, default: '' },
+  },
+  data() {
+    return {
+      hover: false,
+      timeout: null,
```
davidyuk, line 66: Try to avoid timeout by using transitions

src/components/UserHoverCard.vue
davidyuk: Including other places

davidyuk:
I have written an example of how I'm proposing to show Card using Transition and CSS: https://github.com/davidyuk/vue-reproductions/commit/94cfd0a4eca152bcaf72c7dd2950fa763a0cbfd0#diff-c970699f851dadc4a40a1c2a7b29886d5da454125e84e387a8d0abfa8b63be85R1

src/components/UserCard.vue
```diff
@@ -0,0 +1,79 @@
...
+  border-radius: 0.5rem;
+  position: absolute;
+  z-index: 1;
+  overflow: hidden;
+  cursor: default;
+  box-shadow: 4px 4px 4px 0 rgba(0, 0, 0, 0.3);
+  background-color: $thumbnail_background_color;
+}
+
+::v-deep .profile__stats {
```
davidyuk, line 39: Put this inside `.user-card`, otherwise, it looks like a mutation of other component's styles without any condition (like being inside `.user-card`).

src/components/UserCard.vue
```diff
@@ -0,0 +1,79 @@
...
+}
+
+::v-deep .profile__image {
+  .avatar {
+    width: 3.5rem;
+    height: 3.5rem;
+  }
+}
+
+@media screen and (max-width: 1024px) {
```
davidyuk, line 74: Use [mobile mixin](https://github.com/aeternity/superhero-ui/blob/ded59a3317cbb4cb7f51fa78d8db462e879360a3/src/styles/_mixins.scss#L16-L20) instead, and put it inside `.user-card` like [here](https://github.com/aeternity/superhero-ui/blob/1df0e5c8aa9ea6c6df49b091af59ea1ac55585dc/src/views/TipsList.vue#L78-L80).

src/components/tipRecords/AuthorAndDate.vue
```diff
@@ -1,51 +1,71 @@
...
+          :address="address"
+          @mouseover.native="hover = true"
+          @mouseleave.native="hover = false"
+        />
+        <div class="author-name">
+          <span
+            v-if="name"
+            class="chain-name"
+          >
+            {{ name }}
```
davidyuk, line 33: Have you intentionally removed `$t('FellowSuperhero')` from here?

src/components/tipRecords/AuthorAndDate.vue
```diff
@@ -1,51 +1,71 @@
 <template>
-  <div
-    class="author-and-date"
-    @click.stop
```
davidyuk, line 4: Can we revert this `@click.stop` and use it instead of one below and one in UserCard?

src/components/tipRecords/AuthorAndDate.vue
```diff
@@ -1,51 +1,71 @@
...
     address: { type: String, required: true },
     name: { type: String, default: '' },
   },
+  data: () => ({ hover: false }),
 };
 </script>
 
 <style lang="scss" scoped>
-.author-and-date {
+.info {
```
davidyuk, line 68: Put this inside `.author-and-date`

davidyuk:
The overall UI is a bit more flickers than it should be. I'm proposing to avoid that by putting Avatar and UserCard into the same element (to track hover state on it), check 3ee07056edaf05adb20e921f4feb98adec98cdd8.

mradkov (approved):
Looks good to me now.

PR #817, opened 2020-10-01
Don't use event bus in several cases

### LINKED ISSUE #660: Use vuex instead EventBus pattern

davidyuk:
fixes partially #660

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -16,6 +16,15 @@ Vue.config.productionTip = false;
 registerModals();
 sync(store, router);
 
+store.watch(
+  ({ isBackendLive }) => isBackendLive,
+  async (isBackendLive) => {
+    const targetRouteName = isBackendLive ? 'tips' : 'maintenance';
+    if (router.currentRoute.name === targetRouteName) return;
+    await router.push({ name: targetRouteName });
+  },
+);
+
 export default new Vue({
   el: '#app',
   store,
--- a/src/store/index.js
+++ b/src/store/index.js
@@ -2,9 +2,11 @@ import Vue from 'vue';
 import Vuex from 'vuex';
 
 import mutations from './mutations';
+// eslint-disable-next-line import/no-cycle
 import backend from './modules/backend';
 import persistState from './plugins/persistState';
 import modals from './plugins/modals';
+// eslint-disable-next-line import/no-cycle
 import Backend from '../utils/backend';
 
 Vue.use(Vuex);
@@ -32,6 +34,7 @@ export default new Vuex.Store({
     useSdkWallet: false,
     useIframeWallet: false,
     sdk: null,
+    isBackendLive: true,
   },
   mutations,
   actions: {
--- a/src/store/modules/backend.js
+++ b/src/store/modules/backend.js
@@ -3,6 +3,7 @@ import Vue from 'vue';
 import { times } from 'lodash-es';
 // eslint-disable-next-line import/no-cycle
 import Backend from '../../utils/backend';
+import { createDeepLinkUrl } from '../../utils';
 
 export default {
   namespaced: true,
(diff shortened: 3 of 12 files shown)


src/main.js
```diff
@@ -16,6 +16,15 @@ Vue.config.productionTip = false;
...
 sync(store, router);
 
+store.watch(
+  ({ isBackendLive }) => isBackendLive,
+  async (isBackendLive) => {
+    const targetRouteName = isBackendLive ? 'tips' : 'maintenance';
+    if (router.currentRoute.name === targetRouteName) return;
+    await router.push({ name: targetRouteName });
+  },
+);
```
davidyuk, line 26: Don't want to make store and router depend on each other. Can't find a better place for this code 🤔

src/utils/backend.js
```diff
@@ -1,5 +1,31 @@
 import { BACKEND_URL } from '../config/constants';
-import { wrapTry } from './util';
+// eslint-disable-next-line import/no-cycle
+import store from '../store';
+
+const wrapTry = async (promise) => {
```
davidyuk, line 5: Moved this method from utils to don't make the whole utils to depend on the store.

src/views/Maintenance.vue
```diff
@@ -12,25 +12,21 @@ import { EventBus } from '../utils/eventBus';
 export default {
   mounted() {
     const interval = setInterval(() => EventBus.$emit('reloadData'), 10 * 1000);
```
davidyuk, line 14: Actually, it calls only a `reloadData` method of App component. I think it should be extracted to the store and called explicitly.

src/store/index.js
```diff
@@ -2,9 +2,11 @@ import Vue from 'vue';
 import Vuex from 'vuex';
 
 import mutations from './mutations';
+// eslint-disable-next-line import/no-cycle
```
davidyuk, line 5: I propose to leave it for future refactoring. When components will depend on Backend module of the store instead of backend methods directly then we will be able to get rid of cycle dependencies.

Components shouldn't depend on the backend methods file because that file obviously depends on the store (to track backend status), but the components shouldn't depend on the store explicitly for testing/flexibility purposes.

IvayloPapazov (approved):

thepiwo (approved):
looks okay, as far as I understand the changes

CedrikNikita (commented):
```TypeError: Cannot read property 'challenge' of null```
while sending comment without logged in extenstion of iframe wallet

**Review by CedrikNikita — CHANGES_REQUESTED** [review]
Plus actually sent message doesn't show, it'll show afrer reloading the page.

File: src/store/modules/backend.js [hunk 5]
```diff
@@ -110,5 +112,43 @@ export default {
...
+    async sendComment(
+      { dispatch, rootState: { address, useSdkWallet } },
+      { tipId, text, parentId },
+    ) {
+      if (!useSdkWallet) {
+        window.location = createDeepLinkUrl({
+          type: 'comment', id: tipId, text, parentId,
+        });
+        return;
+      }
```
davidyuk, line 140: I have missed this part initially. Bringing it back to fix:
> TypeError: Cannot read property 'challenge' of null

src/store/modules/backend.js
```diff
@@ -110,5 +112,43 @@ export default {
...
+      await dispatch('callWithAuth', {
+        method: 'sendTipComment',
+        arg: {
+          tipId, text, author: address, parentId,
+        },
+      });
+      await Promise.all([
+        dispatch('reloadTip', tipId),
+        ...parentId ? [dispatch('reloadComment', parentId)] : [],
```
davidyuk, line 149: Fetching of both tip and comment fixes updates when replying to a comment on a tip page.

CedrikNikita (approved):
Changes looks reasonable, but this is not compatible with the current production version of the wallet (you can not send a comment via deepLink with this changes).

davidyuk:
Do you mean because there is not applied https://github.com/aeternity/superhero-wallet/pull/571?

PR #819, opened 2020-10-02
Remove event bus from tip input

IvayloPapazov:
part of: #660


IvayloPapazov:
> Why don't extract the whole `sendTip` method?

Because I'm not sure that all variables should be in the store. If you want, I will extract the whole method.

thepiwo:
whats the state of this?

davidyuk:
This one got stuck. I think we need to extract the whole `sendTip` method to store. But now would be better to finish SSR refactoring. Proposing to close this PR and fix the issue later. The link to this code will remain in the connected issue, so these chages may be considered later by one who will be solving the issue.

PR #827, opened 2020-10-07
Remove custom timeout exception generator

davidyuk:
We have issues with custom timeouts, I'm proposing to remove them. Was there any specific reason for them? As I know, HTTP has native timeouts and they should work ok.

### DIFF
--- a/src/utils/util.js
+++ b/src/utils/util.js
@@ -7,23 +7,18 @@ const atomsToAe = (atoms) => (new BigNumber(atoms)).dividedBy(new BigNumber(1000
 const aeToAtoms = (ae) => (new BigNumber(ae)).times(new BigNumber(1000000000000000000));
 export const wrapTry = async (promise) => {
   try {
-    return Promise.race([
-      promise.then((res) => {
-        if (!res) {
-          EventBus.$emit('backendError');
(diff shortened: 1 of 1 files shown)


mradkov:
This was the reason. https://github.com/aeternity/superhero-ui/pull/798#issue-483750940

davidyuk:
It doesn't look like a reason, more like an adjustment of custom timeout introduced before. But why this timeout was introduced?

mradkov:
As the backend is fetching and caching the data from the smart contract it took ~3-4 sec to have its cache reloaded and updated. So initially this was implemented for that reason so the backend has time to update the cache before we reload the page. 🤔

thepiwo:
the timeouts were there to prevent blocking of frontend loading in case the backend is slow, but as we now depend much more on the backend this is maybe not required any more

davidyuk:
@mradkov I'm not proposing to remove the delay introduced in #798 

The thing is that in some cases it breaks crucial stuff
Wire 2020-10-06 at 6_06 PM
Wire 2020-10-06 at 6_08 PM

davidyuk:
the same change got merged in #825, so closing this one

PR #835, opened 2020-10-08
SearchInput: Debounce requests

### LINKED ISSUE #826: Seach input debounce
Let's brainstorm this as the search currently sends a request immediately on each keypress which is not nearly efficient.

ref: http://underscorejs.org/#debounce

davidyuk:
closes #826 
https://lodash.com/docs/4.17.15#debounce
https://css-tricks.com/debouncing-throttling-explained-examples/

### DIFF
--- a/src/components/layout/SearchInput.vue
+++ b/src/components/layout/SearchInput.vue
@@ -35,6 +35,8 @@
 </template>
 
 <script>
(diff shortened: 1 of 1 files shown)


mradkov (approved):
LGTM

PR #839, opened 2020-10-09
Drop superhero-button@0.1.* support

davidyuk:
This was the first version based on iframes, I don't think that somebody still uses it.


mradkov:
Can we assure somehow it is not used anymore? 🤔

mradkov (approved):

PR #832, opened 2020-10-07
fungible token featureflagged via existence of v2 contract

thepiwo:
still waiting for backend changes to be testable

PR #810, opened 2020-09-21
[Prototype] Fungible Tokens & Direct Tipping

stanislav-slavov:
Requires to be used with aeternity/tipping-community-backend#152

This PR supersedes #742. 

Rebased with the latest `develop` branch and removed 'Merge' commits.

PR #842, opened 2020-10-09
Fungible Tokens & Direct Tipping

davidyuk:
the rebased version of #810 with applied #832 
I have squshed the overlapping changes, reducing the commit amount that have simplified rebasing. The diff is almost the same as in the base branch, except that I have extracted #836, you can compare branches by `git diff feature/Fungible-Tokens-Direct-Tipping-3 feature/fungible-token-featureflagged`. I'm going to test this and check the iframe wallet state.


thepiwo:
How does squashing provide a better history here? The history now shows that all changes were made by @davidyuk, when you look at the git blame it shows @davidyuk for all lines changed, so if some questions arise about some lines of code in the future the author will not be known.

thepiwo:
@davidyuk @mradkov I have 2 suggestions
 - keep the non-squashed history to reflect what has happened
 - apply me a author of these squashed commits

davidyuk:
Well, the breaking of the author wasn't intentional, it is because of my git gui, I have fixed that.

> keep the non-squashed history to reflect what has happened

All fixes, adjustments after code review/rebasing are not needed. If you need to fix something after rebasing it means that the rebasing was did wrong. Review remarks usually are quite obvious and don't need to be preserved. So long branch of commits is more difficult to rebase, I think squashing was needed to be done by Stan in the previous PR. The line when history becomes immutable is merging to develop, until that I think we can freely polish it.

thepiwo (approved):

mradkov (approved):

Issue #606, opened 2020-06-04
Sync integrated web-wallet login and extension


mradkov:
The idea is if you have the extension installed - to be still presented the section with the web-wallet, synced with your account credentials.

davidyuk:
I don't think that we need new designs for that. This task is technically complex, but it should be doable. Most probable it required some Vuex refactoring, setting a syncing of state in different directions.

PR #843, opened 2020-10-09
Increase iframe wallet timeout, add a button to enable it manually

davidyuk:
original issue: on #810 web wallet sometimes is opening even if the extension is installed

The increasing of timeout helps in that case, so probably app initialization takes a bit longer on that branch. But initialization time also depends on the hardware and network conditions so, as in https://github.com/aeternity/superhero-wallet/pull/540, I'm proposing to don't depend on timeouts.

In particular, I think we should increase the timeout so some upper bound (10 seconds instead of 2?) and to allow the user to open the web wallet manually if he wants to use it. Until we find a way to use the extension and the web wallet simultaneously (https://github.com/aeternity/superhero-ui/issues/606).

Screenshot 2020-10-09 at 08 53 29

### DIFF
--- a/src/components/layout/FooterSection.vue
+++ b/src/components/layout/FooterSection.vue
@@ -58,10 +58,7 @@
       v-if="!isLoggedIn"
       class="login-footer"
     >
-      <OutlinedButton
-        :to="addressDeepLink"
-        :title="$t('components.layout.FooterSection.LoginWithWallet')"
-      >
+      <OutlinedButton :to="addressDeepLink">
         {{ $t('components.layout.FooterSection.LoginWithWallet') }}
       </OutlinedButton>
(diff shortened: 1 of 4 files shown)


src/components/layout/FooterSection.vue
```diff
@@ -58,10 +58,7 @@
       v-if="!isLoggedIn"
       class="login-footer"
     >
-      <OutlinedButton
-        :to="addressDeepLink"
-        :title="$t('components.layout.FooterSection.LoginWithWallet')"
```
davidyuk, line 63: why do we need title the same as the text on button?

CedrikNikita (approved):

mradkov (approved):

PR #857, opened 2020-10-19
Use v1 contract address from superhero-ui@0.1.15

davidyuk:
To fix the error (on https://develop.aggregator.z52da5wt.xyz/):

> Error: Contract with address ct_2Cvbf3NYZ5DLoaNYAU71t67DdXLHeSXhodkSNifhgd7Xsw28Xd not found on-chain or not active
    at Object.x

The old address is taken from https://github.com/aeternity/superhero-ui/blob/1fd0c075441edbe2988fdfd6f70d0f8985f2984a/src/config/constants.js#L11

Shouldn't it be a hotfix?


Liubov-crypto:
In this pull request works tip on the feed and retip content on the feed.

thepiwo (approved):
right, this was the testnet contract

Issue #833, opened 2020-10-08
Would be guards needed if we use plain links instead of buttons with click handler?


davidyuk:
https://github.com/aeternity/superhero-ui/pull/793/commits/df487ff492ff43ab942bd4a07f1b1136d1879ac9

Probably if we use plain links we won't be needed to write guards like in the commit above, making code simpler and more semantic.

shapkarin:
@davidyuk look like it's not possible to separate links and handlers event inside links. In this case it will fires link navigation and event handler simultaneously. And for example at `src/components/tipRecords/TipComment.vue` when user clicks at tip button it will shows input but next page will be changed by link.
But I can try to avoid or change guards.

davidyuk:
mostly resolved in #860

PR #860, opened 2020-10-22
Don't show cursor at comment page

shapkarin:
@davidyuk as you mentioned after [this conversation](https://github.com/aeternity/superhero-ui/issues/833) I just remove cursor from comment page.

### DIFF
--- a/src/components/tipRecords/TipComment.vue
+++ b/src/components/tipRecords/TipComment.vue
@@ -1,7 +1,11 @@
 <template>
   <div
     class="row position-relative tip-comment"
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -4332,6 +4332,11 @@
           "requires": {
             "bn.js": "^4.11.1"
           }
+        },
```
davidyuk: You have not installed any packages, so not needed to commit changes to package-lock

davidyuk (approved):

package-lock.json
shapkarin: Thanks

PR #863, opened 2020-10-22
PlayButton: Avoid dynamic import

davidyuk:


src/components/PlayButton.vue
```diff
@@ -10,21 +10,24 @@
     />
     <img
       v-else-if="!loading"
-      :src="require(`../assets/button${isPlaying ? 'Pause' : 'Play'}.svg`)"
+      :src="isPlaying ? buttonPause : buttonPlay"
```
davidyuk, line 13: Require should always be called with a constant value, otherwise, the build system has to import all files that technically can be reached by that import. In this case, the difference between bundles ([this](https://feature-avoid-dynamic-import.aggregator.z52da5wt.xyz/report.html), [develop](https://develop.aggregator.z52da5wt.xyz/report.html)) is only 0.4kb, but it may be significant in other cases, so in general, we should avoid it.

mradkov (approved):

PR #864, opened 2020-10-22
Refactor comments

davidyuk:


davidyuk:
@mradkov I still can't ask @Liubov-crypto to review a PR :(

Liubov-crypto:
My comments are duplicated as well as tipping:
Знімок екрана 2020-10-23 о 2 03 22 пп
Знімок екрана 2020-10-23 о 2 12 22 пп
Знімок екрана 2020-10-23 о 2 12 34 пп

CedrikNikita (changes requested):
Trying to send a comment without loggin in in the extension or iframe wallet
```popup.js:41 Error: CommentNew: Invalid arguments```

davidyuk:
@CedrikNikita that issue is reproducible also on develop branch. This is a wallet bug fixed in https://github.com/aeternity/superhero-wallet/pull/571

@Liubov-crypto I have accidentally dropped the code that was removing the duplicate comments, I have reverted it back

Liubov-crypto:
The comment appeared twice with a delay of several seconds (9 sec).
Wire 2020-10-28 at 14-58-37
Wire 2020-10-28 at 14-58-43

The comment thread looks different here: https://feature-refactor-comments.aggregator.z52da5wt.xyz/tip/1375_v1/comment/756

and on development: 
https://develop.aggregator.z52da5wt.xyz/tip/1375_v1/comment/756

onvisions:
As far as I know currently we don't have such logic implemented that would allow us to display "Most Relevant" comments. I guess it's something to do with the reccomendation engine. So our logic is purely related to the moment of posting a comment or reply. 
**Most recent comments** (also a new comment by user) should be **displayed on top**after the specific thread. It doesn't make sense to scroll to bottom to see the most recent comments which is the case now. Also the label says "Latest comments"
 which clearly indicates that newest should be seen first.

Additionally we can add an icon to reverse the display order but **latest first should be the default behaviour**.

davidyuk:
> The comment appeared twice with a delay of several seconds

I can't reproduce this

> The comment thread looks different here

I have fixed over filtering that I have added before

Regarding sorting: I have implemented descending sorting of comment list and ascending sorting of comments replies (because the new comment form placed at the end of them).

Liubov-crypto:
I checked, comments are not duplicated. When I comment on a comment in a thread, that comment is not displayed (click on the arrow to view it). I don't know how critical it is...just a report:

 Знімок екрана 2020-11-05 о 8 20 23 пп
Знімок екрана 2020-11-05 о 8 01 35 пп

davidyuk:
> When I comment on a comment in a thread, that comment is not displayed

This works the same as when commenting on a tip

I think this PR is ready to get merged 🚀

davidyuk:
@CedrikNikita please, aprove this one to be able to merge it

CedrikNikita (approved):
Changes look reasonable

PR #871, opened 2020-11-03
Server side rendering

shapkarin:


package.json
```diff
@@ -30,19 +33,23 @@
     "jdenticon": "^2.2.0",
     "jitsi-iframe-api": "^1.0.0",
     "lodash-es": "^4.17.15",
+    "prerender-spa-plugin": "^3.4.0",
```
davidyuk, line 36: As I understand, this package prerenders pages on the build stage. It can't pretender tip page created by the user after publishing the app. We don't need to use this package.

package.json
```diff
@@ -30,19 +33,23 @@
...
     "tipping-contract": "github:aeternity/tipping-contract#develop",
     "vue": "^2.6.12",
     "vue-i18n": "^8.21.1",
     "vue-router": "^3.4.5",
+    "vue-server-renderer": "^2.6.0",
     "vuex": "^3.5.1",
     "vuex-router-sync": "^5.0.0"
   },
   "devDependencies": {
+    "@akryum/vue-cli-plugin-ssr": "~0.6.0",
```
davidyuk, line 48: Please, investigate the difference between this package and https://www.npmjs.com/package/@uvue/vue-cli-plugin-ssr
https://universal-vue.github.io/docs/guide/#getting-started
https://github.com/universal-vue/uvue

src/App.vue
```diff
@@ -1,228 +1,31 @@
 <template>
```
davidyuk, line 1: Revert changes in this page

src/router/index.js
```diff
@@ -0,0 +1,29 @@
+import Vue from 'vue'
```
davidyuk, line 1: Remove this router and two default components

davidyuk (commented):
What is the current problem with @akryum/vue-cli-plugin-ssr?
I have tried to enable ssr in the example app, it works ok:
https://github.com/davidyuk/vue-reproductions/commit/1474d4ed25d9d0c452e2af9a5c22d7e1b3444c98

package.json
shapkarin: I moved it to the devDependencies

vue.config.js
```diff
@@ -56,6 +58,11 @@ module.exports = {
         return [definitions];
       })
       .end()
+      .plugin('ssr')
+      .tap((options) => ({
+        entry: (target) => `./src/entry-${target}`,
+        ...options[0],
+      }))
```
davidyuk, line 65: This part causing an error:
> Error: args must be an array of arguments

Shouldn't it be handled by @akryum/vue-cli-plugin-ssr?

src/router/index.js
```diff
@@ -0,0 +1,28 @@
+import Vue from 'vue';
+import Home from '../views/Home.vue';
+
+Vue.use(VueRouter);
```
davidyuk, line 4: The next exception is:

> Uncaught ReferenceError: VueRouter is not defined

it isn't intended to be reviewed, don't it?

package.json
shapkarin: Removed, due we decide to use `@akryum/vue-cli-plugin-ssr`

package.json
shapkarin: Done

src/router/index.js
shapkarin: Thanks, looks like I need to pass there a real router.

like 
```js
const router = new VueRouter({})
Vue.use(router);
```

Right?

vue.config.js
shapkarin: Not sure, please check https://vue-cli-plugin-ssr.netlify.app/guide/configuration.html

shapkarin: @davidyuk I guess I got what you mean, I didn't define that this option are about `@akryum/vue-cli-plugin-ssr`

src/router/index.js
shapkarin: @davidyuk I will research more

package.json
davidyuk: So, what is the difference between them except that `@akryum/vue-cli-plugin-ssr` is a bit more popular?

vue.config.js
davidyuk: This is not needed because it is already done on the side of `@akryum/vue-cli-plugin-ssr`

src/router/index.js
davidyuk: > If using a module system (e.g. via vue-cli), import Vue and VueRouter and then call `Vue.use(VueRouter)`.

https://router.vuejs.org/guide/#javascript

src/App.vue
```diff
@@ -59,9 +59,6 @@ export default {
       this.reloadData();
     });
     setInterval(() => this.reloadData(), 120 * 1000);
-    EventBus.$on('backendError', () => this.$route.name !== 'maintenance' && this.$router.push({
-      name: 'maintenance',
-    }).catch((err) => { console.error(err); }));
```
davidyuk, line 64: Seems that you have restored a different version of App.vue 🤷‍♀️

.env
```diff
@@ -3,3 +3,4 @@ VUE_APP_JITSI_URL=meet.aepps.com
 VUE_APP_BACKEND_URL=https://raendom-backend.z52da5wt.xyz
 VUE_APP_CONTRACT_V1_ADDRESS=ct_2AfnEfCSZCTEkxL5Yoi4Yfq6fF7YapHRaFKDJK3THMXMBspp5z
 VUE_APP_CONTRACT_V2_ADDRESS=
+BASE_URL=/
```
davidyuk, line 6: Won't it be defined by Vue cli? When I was adding SSR to the example aepp, I didn't need to do that: https://github.com/davidyuk/vue-reproductions/commit/1474d4ed25d9d0c452e2af9a5c22d7e1b3444c98

davidyuk:
> Almost all is ready - only need to define a route somehow, connect cli plugin and Vue app
Server itself is works

Which route?

> Looks like there is actually needs to define new env variable with root route.

I can't get why you need to define `BASE_URL`

I would like to help you, but I can't start the project because of the obvious error that you have not fixed from the last time:
```bash
$ npm run ssr:serve

> superhero-ui@0.2.1 ssr:serve /Users/denis/Documents/superhero-ui
> vue-cli-service ssr:serve

(node:20850) UnhandledPromiseRejectionWarning: Error: args must be an array of arguments
```

package.json
shapkarin: @davidyuk I messed up with tools because of health. Thanks for notice. I will use https://universal-vue.github.io/docs/guide/#getting-started at the next week and make that ASAP.

.env
shapkarin: @davidyuk cool, so you figure out with that? I will try my best at Monday!

shapkarin:
@davidyuk I mixed up with different tools because of pretty the same name and my sickness. Will do my best at Monday.

vue.config.js
shapkarin: thanks, removed

src/App.vue
shapkarin: OK, revert to the current one from the `develop` branch

.env
shapkarin: @davidyuk thanks, fixed, messed up because of health

davidyuk (commented):
Well, it is still is not working, how can I help? Try to make the backend to don't return "Cannot GET /" by disabling as much as possible of our code. After that continuously revert them back resolving issues that will apperar.

.env
```diff
@@ -5,4 +5,4 @@ VUE_APP_NODE_URL=https://mainnet.aeternity.io
 VUE_APP_EXPLORER_URL=https://mainnet.aeternal.io
 VUE_APP_COMPILER_URL=https://latest.compiler.aepps.com
 VUE_APP_CONTRACT_V1_ADDRESS=ct_2AfnEfCSZCTEkxL5Yoi4Yfq6fF7YapHRaFKDJK3THMXMBspp5z
-VUE_APP_CONTRACT_V2_ADDRESS=
+VUE_APP_CONTRACT_V2_ADDRESS=
```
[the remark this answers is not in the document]
shapkarin: stange, I did config IDE, I will recheck

shapkarin:
@davidyuk can you please share yours suggestion about the current error: `Cannot GET /` ?

davidyuk:
let's continue work in #908

PR #872, opened 2020-11-04
Fix UI issues of UserCard

davidyuk:
Well, the goal to achieve a delay in user card showing using pure css is achievable, but the resulting UI is quite poor. I'm fixing two annoying things that were introduced in #814 because of my guidance.

| Case | Before | After |
|---|---|---|
| Accidental hovering on avatar placing the cursor on the place of card afterward | bug-1-before | bug-1-after |
| Multiple hovering on avatar before the card gets hidden | bug-2-before | bug-2-after |

### DIFF
--- a/src/components/tipRecords/AuthorAndDate.vue
+++ b/src/components/tipRecords/AuthorAndDate.vue
@@ -14,15 +14,15 @@
     >
       <div
         class="avatar-wrapper"
-        @mouseover="hover = true"
-        @mouseleave="hover = false"
+        @mouseover="hoverDebounced = true"
+        @mouseleave="hoverDebounced = false"
       >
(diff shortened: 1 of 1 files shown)


AtanasKrondev (approved):
Nice approach!

mradkov (approved):

Liubov-crypto:
The first case works smoothly, the second - preview on hover disappears if I move the cursor away from the user's avatar.

Запис екрана 2020-11-05 о 7 34 15 пп

davidyuk:
> preview on hover disappears if I move the cursor away from the user's avatar

This should happen but only after 0.5s. My gifs are slowed down probably because of converting from mov.

PR #877, opened 2020-11-16
Visualize Fungible Tokens

stanislav-slavov:
The changes include new dropdown component which allows the usage of custom html to meet the design criteria.

It works on the test network, prior to merge it should be restored to mainnet.

Deisgns: https://nx3594.your-storageshare.de/apps/files/?dir=/P1/Superhero/Superhero.com/Desktop/v2/Tokens%20Website%20UI&fileid=699940 1
Screenshot 2020-11-17 at 16 17 02
Screenshot 2020-11-17 at 16 17 14
Screenshot 2020-11-17 at 16 17 39
Screenshot 2020-11-17 at 16 18 11
Screenshot 2020-11-17 at 16 18 19


src/components/layout/RightSectionWallet.vue
```diff
@@ -21,25 +21,50 @@
...
+          :selected="selectedTokenBalance.token"
+        >
+          <template #diplayValue>
+            <AeAmount
+              class="not-bootstrap-row"
+              :amount="selectedTokenBalance.balance"
+              :token="selectedTokenBalance.token"
+            />
+          </template>
+          <template slot-scope="{ option }">
```
davidyuk, line 37: `slot-scope` is deprecated, let's use a new syntax instead
https://vuejs.org/v2/guide/components-slots.html#Scoped-Slots-with-the-slot-scope-Attribute

src/components/Dropdown.vue
```diff
@@ -52,50 +63,62 @@ export default {
...
-    position: relative;
-    color: $standard_font_color;
-    font-size: 0.75rem;
-    background-color: $light_color;
-    padding: 0.6rem 2.2rem 0.6rem 0.85rem;
-    line-height: 0.9rem;
-    min-height: 2.2rem;
+  button {
+    background-color: transparent;
+    border: none;
```
davidyuk, line 70: Try to use ButtonPlain instead

src/components/TipInput.vue
```diff
@@ -13,14 +13,23 @@
           v-if="!tipUrlStats.tokenTotalAmount.length || +tipUrlStats.totalAmountAe !== 0"
           :amount="tipUrlStats.totalAmountAe"
         />
+      </template>
+    </Component>
+    <Dropdown
+      v-if="tipUrlStats && tipUrlStats.tokenTotalAmount.length"
```
davidyuk, line 19: Shouldn't we hide this dropdown on UserProfile page?

src/components/TipInput.vue
```diff
@@ -13,14 +13,23 @@
...
+      :options="tipUrlStats.tokenTotalAmount"
+      show-right
+      class="token-tips"
+      @click.stop
```
davidyuk, line 23: It is unclear in which case it is needed, can't it be done by a parent component?

src/components/layout/RightSectionWallet.vue
```diff
@@ -57,39 +82,65 @@ import AeAmount from '../AeAmount.vue';
...
+    aeternityTokenData() {
+      return {
+        balance: this.balance,
+        token: '',
+      };
+    },
+    tokenBalancesOptions() {
+      // Aeternity token + FT data
+      return [
+        ...[this.aeternityTokenData],
```
davidyuk, line 121:
```suggestion
        this.aeternityTokenData,
```

src/components/layout/RightSectionWallet.vue
```diff
@@ -57,39 +82,65 @@ import AeAmount from '../AeAmount.vue';
...
+        .toFixed(2);
+    },
+    aeternityTokenData() {
+      return {
+        balance: this.balance,
+        token: '',
+      };
+    },
+    tokenBalancesOptions() {
+      // Aeternity token + FT data
```
davidyuk, line 119:
```suggestion
```
This thing should be self-explanatory

src/components/layout/RightSectionWallet.vue
```diff
@@ -21,25 +21,50 @@
...
+        </Dropdown>
+        <AeAmount
+          v-else
+          :amount="balance"
+        />
         <Dropdown
           v-if="currencyDropdownOptions.length"
           :options="currencyDropdownOptions"
-          :method="updateCurrency"
+          :method="updateSelectedCurrency"
```
davidyuk, line 48:
```suggestion
          :method="({ currency }) => updateCurrency(currency)"
```
No need to wrap `updateCurrency` with an additional method for that

src/components/layout/RightSectionWallet.vue
```diff
@@ -21,25 +21,50 @@
...
+          v-if="hasContractV2Address"
+          :options="tokenBalancesOptions"
+          :method="updateTokenBalance"
+          :selected="selectedTokenBalance.token"
+        >
+          <template #diplayValue>
+            <AeAmount
+              class="not-bootstrap-row"
+              :amount="selectedTokenBalance.balance"
+              :token="selectedTokenBalance.token"
```
davidyuk, line 34: These values should be taken as a current dropdown value from the slot scope, no need to define `selectedTokenBalance` for that.

src/components/layout/RightSectionWallet.vue
```diff
@@ -21,25 +21,50 @@
         {{ address }}
       </div>
       <div class="not-bootstrap-row">
-        <AeAmount :amount="balance" />
+        <Dropdown
+          v-if="hasContractV2Address"
+          :options="tokenBalancesOptions"
+          :method="updateTokenBalance"
```
davidyuk, line 27: I can't find that it actually changes except the Dropdown value.

davidyuk (changes requested):

.env
```diff
@@ -1,8 +1,8 @@
 VUE_APP_WALLET_URL=https://wallet.superhero.com
 VUE_APP_JITSI_HOST=meet.aepps.com
-VUE_APP_BACKEND_URL=https://raendom-backend.z52da5wt.xyz
-VUE_APP_NODE_URL=https://mainnet.aeternity.io
-VUE_APP_EXPLORER_URL=https://mainnet.aeternal.io
+VUE_APP_BACKEND_URL=https://test-tipping.aeternity.art
+VUE_APP_NODE_URL=https://testnet.aeternity.io
+VUE_APP_EXPLORER_URL=https://testnet.aeternal.io
```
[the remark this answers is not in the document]
stanislav-slavov: Yes, the last commit would be removed prior to merging the PR, we have to properly test it.

src/utils/aeternity.js
```diff
@@ -76,12 +76,13 @@ export const scanForWallets = async () => {
     connectionInfo: { id: 'spy' },
   });
   const detector = await Detector({ connection: scannerConnection });
-  const webWalletTimeout = setTimeout(() => !IS_MOBILE_DEVICE && store.commit('enableIframeWallet'), 10000);
+  // const webWalletTimeout = setTimeout(() =>
+  // !IS_MOBILE_DEVICE && store.commit('enableIframeWallet'), 10000);
```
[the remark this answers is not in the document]
stanislav-slavov: No, there are no issues with the iframe wallet, but we need to visualize the currency dropdown + the token amount dropdown as per the design. In order to test it properly these changes could be removed prior to merge.

src/components/fungibleTokens/TokenAvatarAndSymbol.vue
```diff
@@ -0,0 +1,56 @@
+<template>
+  <div class="token">
+    <Avatar
+      v-if="tokenData"
+      class="token-avatar"
+      :address="address"
+      show-ident-icon-only
+    />
+    <div
+      v-else
```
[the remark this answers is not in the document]
stanislav-slavov: The aeternity coin logo is constructed with the < div > background 
Screenshot 2020-11-23 at 11 04 14
I don't think I should change it to image 
Screenshot 2020-11-23 at 11 05 38

The class "default" is replaced by "aeternity-logo".

src/components/AeAmount.vue
```diff
@@ -17,6 +23,7 @@ export default {
     round: { type: Number, default: 2 },
     aettos: { type: Boolean, required: false },
     token: { type: String, default: null },
+    withoutSymbol: { type: Boolean },
```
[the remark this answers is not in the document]
stanislav-slavov: The component is called "AeAmount" and is mainly used for formatting balances for Ae and FT. I don't see it as an overhead to add "withoutSymbol" flag.
Adding a getter would mean to go through all balances and find the particular one we need. This should be applied for every token in a list. 
I don't see the need to double this logic:
Screenshot 2020-11-23 at 16 38 03

src/components/TipInput.vue
stanislav-slavov: I guess we should be able to send fungible tokens to user profiles also. @mradkov What do you think?

stanislav-slavov: This is a good question, it is not included in the designs so I will hide it for now.

stanislav-slavov: I think it could go into the "stats" section.

src/components/TipInput.vue
stanislav-slavov: It wasn't actually needed in this particular instance.

src/components/layout/RightSectionWallet.vue
stanislav-slavov: Yeah, by design this dropdown is just displaying the token balance. It is about to be transformed into a search component in near future.

src/components/fungibleTokens/TokenAvatarAndSymbol.vue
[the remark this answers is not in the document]
davidyuk: No, you should. Just add `object-fit: contain` and some padding to the img and it will look like the current two backgrounds approach.

src/components/AeAmount.vue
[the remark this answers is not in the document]
davidyuk: > I don't see the need to double this logic

Don't need to duplicate it, just extract it and reuse it here, and where else it is needed

src/components/TipInput.vue
```diff
@@ -13,14 +13,22 @@
...
         />
+      </template>
+    </Component>
+    <Dropdown
+      v-if="!userAddress && tipUrlStats && tipUrlStats.tokenTotalAmount.length"
+      :options="tipUrlStats.tokenTotalAmount"
+      show-right
+      class="token-tips"
+    >
+      <template v-slot="{ option }">
```
davidyuk, line 24: I have checked it locally and seems that you don't need an extra template:
```suggestion
      v-slot="{ option }"
    >
```

src/components/TipInput.vue
```diff
@@ -13,14 +13,22 @@
           v-if="!tipUrlStats.tokenTotalAmount.length || +tipUrlStats.totalAmountAe !== 0"
           :amount="tipUrlStats.totalAmountAe"
         />
+      </template>
```
davidyuk, line 16: Actually, this template contains only one child node. Please, remove it by applying `v-if` to AeAmountFiat node.

src/components/TipInput.vue
```diff
@@ -236,5 +252,19 @@ export default {
       }
     }
   }
+
+  .token-tips {
```
davidyuk, line 256: In TipInput there is only one dropdown, why don't refer to it by `.dropdown`?

src/components/TipInput.vue
```diff
@@ -236,5 +252,19 @@ export default {
...
+    display: inline-block;
+    top: -0.15rem;
```
davidyuk, line 258: Isn't there any more fancy way to align the button vertically?

src/components/TipInput.vue
```diff
@@ -236,5 +252,19 @@ export default {
       }
     }
   }
+
+  .token-tips {
+    display: inline-block;
```
davidyuk, line 257: I propose to move this to the Dropdown component because until opened it behaves like buttons that are inline-block by default.

src/components/TipInput.vue
```diff
@@ -236,5 +252,19 @@ export default {
...
+    top: -0.15rem;
+    position: relative;
```
davidyuk, line 259: This style is already applied in the Dropdown component

src/components/TipInput.vue
```diff
@@ -236,5 +252,19 @@ export default {
...
+
+    ::v-deep button {
+      padding: 0;
```
davidyuk, line 262: Can it be moved to Dropdown component?

src/components/TipInput.vue
```diff
@@ -236,5 +252,19 @@ export default {
...
+    }
+
+    ::v-deep .not-bootstrap-modal-content {
+      min-width: 9rem;
```
davidyuk, line 266: Does it make sense from the design perspective? Maybe @onvisions just forgot to resize the dropdown to the size of the content.

I'm talking about the difference between:

Screenshot 2020-11-25 at 10 17 29
Screenshot 2020-11-25 at 10 17 45

Stuff like this brings a lot of extra styles that are difficult to maintain.

src/components/TipInput.vue
```diff
@@ -217,6 +228,11 @@ export default {
       min-width: 16rem;
       padding: 0.5rem;
     }
+
+    .not-bootstrap-modal-content {
```
davidyuk, line 232: I can't find `.not-bootstrap-modal-content` inside `.not-bootstrap-modal-content` inside `.not-bootstrap-modal`. Is this styles used somewhere?

src/components/AeAmount.vue
```diff
@@ -1,8 +1,13 @@
 <template>
-  <span class="ae-amount">
+  <span
+    class="ae-amount"
+  >
```
davidyuk, line 4: Please revert this change

src/components/layout/RightSectionWallet.vue
```diff
@@ -57,39 +82,55 @@ import AeAmount from '../AeAmount.vue';
...
+    tokenBalancesOptions() {
+      return [
+        this.aeternityTokenData,
+        ...this.tokenBalances,
+      ];
+    },
+  },
+  methods: {
+    ...mapMutations(['updateCurrency', 'enableIframeWallet']),
   },
```
davidyuk, line 127:
```suggestion
  methods: mapMutations(['updateCurrency', 'enableIframeWallet']),
```
Don't change this without need

davidyuk (changes requested):

src/components/Dropdown.vue
```diff
@@ -1,49 +1,65 @@
...
+    <Modal
+      v-if="showMenu"
+      @close="showMenu = false"
+    >
+      <Component
+        :is="method ? 'button' : 'div'"
+        v-for="option in options"
+        :key="option.value"
+        class="dropdown-item"
+        @click="method ? (method(option), currentValue = option) : null, showMenu = false"
```
davidyuk, line 26: This handler is overcomplicated, please define a separate method instead

src/components/Dropdown.vue
```diff
@@ -52,50 +68,63 @@ export default {
...
+
+  .dropdown-item {
+    font-size: 0.75rem;
+    min-width: 6.5rem;
+    padding: 0.6rem 0.8rem;
+    background-color: $actions_ribbon_background_color;
+    color: $standard_font_color;
+    box-shadow: inset 0 0 0.1rem $article_content_color;
+    border-radius: unset;
+    outline: none;
```
davidyuk, line 94: You can remove this by using ButtonPlain

src/components/CustomDropdown.vue
```diff
@@ -0,0 +1,125 @@
...
+        name="diplayValue"
+      />
+      <span v-if="!$slots.diplayValue">{{ displayValue }}</span>
+      <img src="../assets/carretDown.svg">
+    </button>
+    <Modal
+      v-if="showMenu"
+      @close="showMenu = false"
+    >
+      <div
```
davidyuk: I would like to read commits one by one, but I can't do so because it doesn't contain fixes that you have implemented later.

src/components/fungibleTokens/TokenAvatarAndSymbol.vue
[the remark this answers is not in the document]
stanislav-slavov: I don't think there is a problem with that approach, but ok...

src/components/CustomDropdown.vue
stanislav-slavov: Yes, the fixes are after the dropdown is renamed.

src/components/TipInput.vue
stanislav-slavov: Yes, the dropdown contains a < Modal > component and it is within one so it is a one time case where we have dropdown within a dropdown. This style overrides the parent one.

src/components/fungibleTokens/TokenAvatarAndSymbol.vue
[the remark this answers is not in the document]
stanislav-slavov: I've tried it, the image doesn't scale good. My approach is good, I don't see a legit reason not to follow it...

src/components/Dropdown.vue
davidyuk: @stanislav-slavov clean up the styles, `outline: none` is still there

src/components/fungibleTokens/TokenAvatarAndSymbol.vue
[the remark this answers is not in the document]
davidyuk: I agree that it is not important, but would be nice if you know how to position the image this way, check the commit that I have added to your branch a6501c0c564d673c17176ea09dc0abfc97de59cc

src/components/TipInput.vue
davidyuk: Ok, I got that. But, actually, you can remove this block by making the selector more specific as I did in 2311feb04d35a81c7aa07b201903f5dd223420c6

src/components/Dropdown.vue
stanislav-slavov: Removed it, thanks.

src/components/fungibleTokens/TokenAvatarAndSymbol.vue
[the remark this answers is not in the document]
stanislav-slavov: Thanks.

src/components/TipInput.vue
davidyuk: > Tsvetan Kolev
You have posted there 2 identical screenshots as far as I see

@stanislav-slavov so I think we can ignore predefined widths in similar cases.

stanislav-slavov: Ok, it is removed from the instance.

onvisions:
I didn't forget to resize the dropdown according to content as I intended to set here its **minimum width** to be equal to the width of the tips display element (including the diamond icon). @davidyuk is right that it's not so important and we can have dropdown width defined just by the size of its content. Anyway we should have cross-browser consistency regarding the padding. Meaning that with the incoming rebranding and UI improvement we will be using 8pt grid layout so the padding I suggest here are: 8px 16px 8px 16px for each item of the dropdown. So with auto width it will look like this:
Group 1617

Liubov-crypto:
Tested - works well on my side.

src/components/__tests__/AeAmount.js
```diff
@@ -1,5 +1,15 @@
...
+import Vuex from 'vuex';
 import AeAmount from '../AeAmount.vue';
+import Getters from '../../store/getters';
+
+const localVue = createLocalVue();
+
+localVue.use(Vuex);
+
+const store = new Vuex.Store({
+  getters: Getters,
```
davidyuk, line 11: Now it looks more like an integration test than a unit test. I think the idea is to stub `roundedTokenAmount` with some simple function, test exactly AeAmount, and then test `roundedTokenAmount` in a separate file.

src/components/layout/RightSectionWallet.vue
```diff
@@ -21,25 +21,50 @@
...
         <Dropdown
-          v-if="currencyDropdownOptions.length"
+          v-if="hasContractV2Address"
+          :options="tokenBalancesOptions"
+          :method="option => showCurrencyDropdown = !(option.token.length > 0)"
+          :selected="aeternityTokenData.token"
+        >
+          <template #displayValue="{ displayValue }">
+            <AeAmount
+              class="not-bootstrap-row"
```
davidyuk, line 32:
```suggestion
```
this doesn't seem to be needed

src/components/layout/RightSectionWallet.vue
```diff
@@ -21,25 +21,50 @@
...
           :options="currencyDropdownOptions"
-          :method="updateCurrency"
+          :method="({ currency }) => updateCurrency(currency)"
           :selected="selectedCurrency"
-          rounded
-        />
+          show-right
+        >
+          <template #displayValue>
+            <span class="currency-value spaced">
```
davidyuk, line 53: the `spaced` is only needed because, for some reason, the Dropdown displays its value inside of the flexbox

davidyuk (approved):
Well, I'm giving up. Here is the last remarks. This code can be polished further, but it takes too much time🤷‍♀️

src/components/layout/RightSectionWallet.vue
stanislav-slavov: The value could be anything(thats how I wanted it to work). The styling is supposed to be done when you pass the slot and customize it.

Liubov-crypto:
FT operations are now faster
I see FT a few seconds after retipping content on the feed.

Liubov-crypto:
There is an issue with re-tipping some content (V1) using FT in superhero feed.
I sent 0.1 AMZ to https://feature-fungible-tokens-ui.aggregator.z52da5wt.xyz/tip/28_v1. 
The FTs were debited from my wallet, but I don't see them in the amount of tokens post I re-tiped: https://testnet.aeternal.io/transactions/th_AYJCSjXfKGq3ogFqUg5TYF8W9D4a9CV8S863vRTTprpEkvR4N

There were also errors with re-tipping other V1 posts: 
error in chrome
error in chrome1

stanislav-slavov:
Removed testnet configurations. The PR could be merged.

mradkov (approved):

PR #895, opened 2020-11-25
Extract text from Venture logo

davidyuk:
While #894 I have noticed that that logo's font is broken. According to SVG, the font should be Roboto, 11px. It is the same as the near text, so I assumed that is should be fine to extract "Branded by Venture" as text ant to put it near their logo.

Before:

Screenshot 2020-11-25 at 07 05 05

After:

Screenshot 2020-11-25 at 07 05 09


mradkov (approved):
Nice catch. Changes look reasonable.

PR #899, opened 2020-11-27
Dropdown: Remove data wrapper around prop

CedrikNikita:
fixes https://github.com/aeternity/superhero-wallet/issues/660

### DIFF
--- a/src/components/Dropdown.vue
+++ b/src/components/Dropdown.vue
@@ -16,7 +16,7 @@
         @change="method(selectedVal)"
       >
         <option
-          v-for="(option, idx) in optionsVal"
+          v-for="(option, idx) in options"
           :key="idx"
           :value="option.value"
         >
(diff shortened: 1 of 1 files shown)


src/components/layout/RightSectionWallet.vue
```diff
@@ -23,7 +23,7 @@
       <div class="not-bootstrap-row">
         <AeAmount :amount="balance" />
         <Dropdown
-          v-if="currencyDropdownOptions.length"
+          v-if="currencyDropdownOptions.length && balance"
```
davidyuk, line 26: Would be nice to hide the dropdown in this case, but it is not related to #660 bug because obviously on the screenshots from there the balance is not 0, but the dropdown contains 0 currency amounts.

Liubov-crypto:
Checked - looks good.

davidyuk (approved):

PR #901, opened 2020-11-27
refactor components, extracted from wordbazaar changes

thepiwo:


src/App.vue
```diff
@@ -93,8 +93,17 @@ export default {
...
-          await Promise.all(Object.entries(tokens).map(async ([token]) => this
-            .addTokenBalance({ token, balance: await tokenBalance(token, this.address) })));
+          await Promise.all(Object.entries(tokens).map(async ([token]) => {
+            this.addTokenBalance({
+              token,
+              balance: await tokenBalance(token, this.address),
+            });
+            this.addTokenPrice({
+              token,
+              price: await Backend.getWordSaleDetailsByToken(token)
```
thepiwo, line 103: will always fail and fallback to 0, but thats okay for now

src/components/AeButton.vue
```diff
@@ -51,5 +51,9 @@ export default {
   span {
     vertical-align: inherit;
   }
+
+  &.green {
```
davidyuk, line 55: I would switch this class by a boolean property instead because seeing it's usage it can be mixed up with:
- a class to add styles locally (maybe accidentally removed on refactoring as unused)
- some global class that makes anything green

src/components/BackButtonRibbon.vue
```diff
@@ -1,14 +1,27 @@
...
   <div class="actions-ribbon">
-    <img
-      src="../assets/backArrow.svg"
-      @click="back"
-    >
+    <div class="back-button">
+      <img
+        src="../assets/backArrow.svg"
+        @click="back"
+      >
```
davidyuk, line 7:
```suggestion
    <ButtonPlain
      class="back-button"
      @click="back"
    >
      <img src="../assets/backArrow.svg">
```

src/components/FiatValue.vue
```diff
@@ -10,17 +10,39 @@
...
+      const shiftBy = this.token ? -this.tokenData.decimals : shiftAettos;
+      const price = this.token ? this.tokenData.price : 1;
+
+      return new BigNumber(shiftDecimalPlaces(this.amount, shiftBy))
+        .multipliedBy(price).multipliedBy(this.rate)
+        .toNumber()
+        .toLocaleString('en-US', { style: 'currency', currency: selectedCurrency });
+    },
+    tokenData({ tokenPrices, tokenInfo }) {
+      console.log(this.token, tokenPrices, tokenInfo);
```
davidyuk, line 38: ✂️

src/components/ListOfTipsAndComments.vue
```diff
@@ -1,42 +1,18 @@
...
-        :class="{ active: activeTab === 'comments' }"
-        @click="setActiveTab('comments')"
-      >
-        {{ $t('comments') }}
-      </a>
-    </div>
+    <ActivityRibbon
+      :tabs="ribbonTabs"
+      :activity="activity"
+      :set="(setActivity) => activity = setActivity"
```
davidyuk, line 6:
```suggestion
      v-model="activity"
```
🤷‍♀️

src/components/Loading.vue
```diff
@@ -1,15 +1,15 @@
 <template>
   <div
     class="loading"
-    :class="{ 'above-content': aboveContent }"
+    :class="{ 'above-content': aboveContent, 'small': small }"
```
davidyuk, line 4:
```suggestion
    :class="{ 'above-content': aboveContent, small }"
```

src/components/TabBar.vue
```diff
@@ -0,0 +1,53 @@
+<template>
+  <div class="tab-bar">
+    <a
```
davidyuk, line 3:
```suggestion
    <ButtonPlain
```
would be accurate to use it because, actually, this is not a link

src/store/mutations.js
```diff
@@ -56,6 +56,11 @@ export default {
       .filter((bal, idx, arr) => !arr.slice(0, idx).some((b) => b.token === bal.token))
       .sort((a, b) => a.token.localeCompare(b.token));
   },
+  addTokenPrice(state, payload) {
+    const { tokenPrices } = state;
+    tokenPrices[payload.token] = payload.price;
+    state.tokenPrices = tokenPrices;
```
davidyuk, line 62:
```suggestion
  addTokenPrice({ tokenPrices }, payload) {
    Vue.set(tokenPrices, payload.token, payload.price);
```
`Vue.set` will trigger reactive updates

davidyuk (approved):
In general, the code looks good. Would be nice to implement some of suggestions, at least ones that don't make breaking changes towards wordbazaar branch.

src/App.vue
```diff
@@ -93,8 +93,17 @@ export default {
         })(),
         (async () => {
           const tokens = await Backend.getTokenBalances(this.address);
-          await Promise.all(Object.entries(tokens).map(async ([token]) => this
-            .addTokenBalance({ token, balance: await tokenBalance(token, this.address) })));
+          await Promise.all(Object.entries(tokens).map(async ([token]) => {
```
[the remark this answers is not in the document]
thepiwo: how do you mean?

src/components/ListOfTipsAndComments.vue
thepiwo: this didn't seem to work, don't know why

src/components/AeButton.vue
thepiwo: just currently unused, will be used in word bazaar

is local to that class so should be fine from a global view

src/components/ListOfTipsAndComments.vue
thepiwo: figured it out

src/App.vue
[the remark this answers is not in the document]
thepiwo: currently none of the backend call resulting data is persisted, which should be for the overall logic, just moving the call to a different store doesn't help

also as we rely on data from aeternity as well I don't think it should be in store

src/components/AeButton.vue
davidyuk: I mean, it is unclear when it is used like:
```vue
<template>
  <AeButton class="green" />
</template>

<style scoped>
.red {
  color: red;
}
</style>
```
- is green defined globally?
- is green had corresponding styles defined in this component before?
- is green defined inside AeButton?

Using property it becomes obvious that `green` is defined on the side of AeButton:
```vue
<template>
  <AeButton green />
</template>
```

> just currently unused, will be used in word bazaar

it may look unused like the definition is missed like in the example above

> is local to that class so should be fine from a global view

it doesn't pollute the global scope, but its usage may look like it is defined global as in the example above

After all, it is not consistent with some other components like [Loading](https://github.com/aeternity/superhero-ui/pull/901/files#diff-7fb57159f993f8d56e574f7be6ea421b0cdb807a44881a10ea4022d352061f04R12).

src/App.vue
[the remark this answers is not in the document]
davidyuk: > currently none of the backend call resulting data is persisted, which should be for the overall logic, just moving the call to a different store doesn't help

Firstly, App is not a good place to store a blob of initialization logic that is needed for other components. It is ok to trigger this initialization from here. Probably the best could be to do a self-contained plugin for the store that will initialize itself when it is called by the store.

The question is not in persisting, I'm proposing to keep together the definition of the state, it's mutations, and actions (like fetching of token data), and to clean up App at the same time.

> also as we rely on data from aeternity as well I don't think it should be in store

In general, we should have
- a set of components depending on its internal state and vuex
- internal state and vuex be mutated by methods or mutations called by actions
- helper functions that may work with some API or do utility functions, but they shouldn't depend on the vuex

I think the whole "aeternity.js" should be in vuex. Because
- it is the common state and methods that are needed for multiple components (can't be a part of a single component)
- it depends on the store's state (sdk, current network), so makes the explicit dependency on the store.

Making the explicit dependency between components and the store:
- complicates unit testing of components (more difficult to stub the store)
- complicates SSR (the component become dependant on a single instance of a store that may be multiple in a single server thread)

Check 5ae9ce28a98cb28600b18176db3b47d721de8dc8 for the way how I'm proposing to move tokens data fetching to store.

mradkov:
@thepiwo Can you rebase with develop and fix the 3 conflicts so we can merge it?

src/components/AeButton.vue
thepiwo: Sorry I didn't understand earlier, thanks for the explanation, adjusted here https://github.com/aeternity/superhero-ui/pull/901/commits/6fc8159178229c11a2ef09c9a13655be5e828bcd

davidyuk (approved):

mradkov (approved):

PR #902, opened 2020-11-30
Tip image CSS refactoring.

### LINKED ISSUE #873: Use background-position instead of absolute positioned img in tip previews

shapkarin:
Closes #873


shapkarin:
@davidyuk please, check this kind of approach.

shapkarin:
@davidyuk @mradkov please have a look.

| Before  | After |
| ------------- | ------------- |
|  |  |
| |   |

shapkarin:
@davidyuk @mradkov please have a look. I think that `After` (the current PR) is looks better.

| Before  | After |
| ------------- | ------------- |
|  |  |
|  |   |
|  |  |
|  |  |

davidyuk:
Screenshot 2020-12-02 at 11 31 06
the implementation in this PR allows to brings more attention by using a taller images, this is not a healthy thing

src/components/tipRecords/TipPreview.vue
```diff
@@ -388,28 +388,20 @@ export default {
...
-    display: block;
-    left: 0;
-    margin: auto;
     max-width: 100%;
     width: 100%;
-    position: absolute;
-    right: 0;
-    top: 0;
     border: none;
     height: auto;
```
davidyuk, line 402:
```suggestion
    height: 319px; //?
    object-fit: cover/contain;
```
instead of absolute positioning

davidyuk (commented):
Proposing to don't make significant changes in this PR

thepiwo:
whats the state of this?

shapkarin:
@thepiwo actually it's done pretty well I guess.
@mradkov can you please check this PR?

src/components/tipRecords/TipPreview.vue
```diff
@@ -388,28 +388,21 @@ export default {
...
-    margin: auto;
     max-width: 100%;
     width: 100%;
-    position: absolute;
-    right: 0;
-    top: 0;
     border: none;
-    height: auto;
+    height: 319px; //?
+    object-fit: cover/contain;
```
davidyuk, line 403: This is not a valid css, I mean to use one of them

davidyuk (changes requested):

src/components/tipRecords/TipPreview.vue
shapkarin: thanks

shapkarin: fixed

davidyuk:
Please, rebase it on the top of develop. What was the reasoning for making visual changes? In my issue, I meant only to simplify styles. I'm not in charge to decide how previews should look like, from my perspective in some cases this PR makes them better in some worse:

Screenshot 2021-02-22 at 10 26 42
Screenshot 2021-02-22 at 10 28 45

From the code perspective, you are adding the hard-coded height two times that doesn't look good. Why don't remove the img node and use the background property as I suggested initially? It looks like that you are overcomplicating a simple intention.

shapkarin:
@davidyuk There was discussion about such change with @mradkov 
didn't check the last update..
maybe better is to use `contain`?

shapkarin:
@davidyuk I guess it's better to use `contain`. Looks like now it's a bit better look.

davidyuk:
@mradkov what do you think?

PR #905, opened 2020-12-02
Fix retiping through deepLink

CedrikNikita:


mradkov (approved):

src/components/TipInput.vue
```diff
@@ -122,7 +122,7 @@ export default {
     },
     deepLink() {
       return createDeepLinkUrl(this.tip
-        ? { type: 'retip', id: this.tip.id } : { type: 'tip', url: this.tipUrl });
+        ? { type: 'retip', id: this.tip.id.split('_')[0] } : { type: 'tip', url: this.tipUrl });
```
davidyuk, line 125: So, the contract version should be detected by wallet then? This is not consistent, with, for example, comments because there we already passing the complete id there (including version). Also, the same bug is fixed on the wallet side in https://github.com/aeternity/superhero-wallet/pull/682

CedrikNikita: I missed that part, yeah we may fix it on the wallet side, i agree

PR #906, opened 2020-12-06
Update favicon config

davidyuk:
Generated Firefox app icon

Before 😆  | After
---|---
firefox_app_512x512 | firefox_app_512x512

### DIFF
--- a/vue.config.js
+++ b/vue.config.js
@@ -25,21 +25,17 @@ module.exports = {
       .plugin('favicons')
       .use(FaviconsWebpackPlugin, [{
         logo: path.resolve(__dirname, 'src/assets/favicon.svg'),
-        mode: 'webapp', // optional can be 'webapp' or 'light' - 'webapp' by default
-        devMode: 'webapp', // optional can be 'webapp' or 'light' - 'light' by default
         publicPath: '/',
         favicons: {
           start_url: '/',
-          appName: 'You Do Not Need A Cape To Be A Hero',
-          appDescription: 'You Do Not Need A Cape To Be A Hero',
+          appName: 'Superhero',
+          appDescription: 'Social platform with a mission',
           developerName: 'Aeternity Developers',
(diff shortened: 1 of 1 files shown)


vue.config.js
```diff
@@ -25,22 +25,15 @@ module.exports = {
       .plugin('favicons')
       .use(FaviconsWebpackPlugin, [{
         logo: path.resolve(__dirname, 'src/assets/favicon.svg'),
-        mode: 'webapp', // optional can be 'webapp' or 'light' - 'webapp' by default
-        devMode: 'webapp', // optional can be 'webapp' or 'light' - 'light' by default
```
davidyuk, line 29: The default values should be fine. I don't think that it is necessary to generate `2732x2048` images on each dev run.

vue.config.js
```diff
@@ -25,22 +25,15 @@ module.exports = {
...
         publicPath: '/',
         favicons: {
           start_url: '/',
-          appName: 'You Do Not Need A Cape To Be A Hero',
+          appName: 'Superhero.com',
           appDescription: 'You Do Not Need A Cape To Be A Hero',
```
davidyuk, line 32: Probably this is an outdated description.

vue.config.js
```diff
@@ -25,22 +25,15 @@ module.exports = {
...
           developerName: 'Aeternity Developers',
-          developerURL: 'https://github.com/kenodressel/aepp-boilerplate',
-          background: '#ff0d6a',
```
davidyuk, line 37: The colors from Base aepp 😅

vue.config.js
```diff
@@ -25,22 +25,15 @@ module.exports = {
...
-          theme_color: '#ff0d6a',
-          icons: {
-            coast: false,
-            yandex: false,
-            windows: false,
-          },
```
davidyuk, line 43: I can't find an explanation of why it is disabled in git blame. I think it is ok to use all features implemented in this package.

CedrikNikita (approved):

PR #907, opened 2020-12-06
Enable SSR using @akryum/vue-cli-plugin-ssr

davidyuk:
Currently, it looks this way with disabled JS:

Screenshot 2020-12-06 at 17 25 09

The missing part is the fetching of the tip list/specific tips on the server-side.

Also, I'm going to try `@uvue/vue-cli-plugin-ssr` because it is still in active development.

### DIFF
--- a/vue.config.js
+++ b/vue.config.js
@@ -9,6 +9,11 @@ const commitHash = require('child_process')
 
 module.exports = {
   lintOnSave: false,
+  pluginOptions: {
+    ssr: {
+      favicon: './dist/assets/favicon.ico',
+    },
+  },
   css: {
     loaderOptions: {
       sass: {
--- a/vue.config.js
+++ b/vue.config.js
@@ -12,7 +12,7 @@ module.exports = {
   pluginOptions: {
     ssr: {
       favicon: './dist/assets/favicon.ico',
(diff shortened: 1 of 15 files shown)


vue.config.js
```diff
@@ -9,6 +9,12 @@ const commitHash = require('child_process')
 
 module.exports = {
   lintOnSave: false,
+  pluginOptions: {
+    ssr: {
+      favicon: './dist/assets/favicon.ico',
```
davidyuk, line 14: The "Cannot GET /" error was because of this https://github.com/Akryum/vue-cli-plugin-ssr/blob/e9d0c4fc448e11805edde294d2cb9654c1daa3ce/lib/app.js#L94
is was caused by a wrong path to favicon.

vue.config.js
```diff
@@ -9,6 +9,12 @@ const commitHash = require('child_process')
...
+      nodeExternalsWhitelist: [/\.css$/, /\?vue&type=style/, 'lodash-es'],
```
davidyuk, line 15: This is based on https://github.com/Akryum/vue-cli-plugin-ssr/issues/45#issuecomment-452719512 and fixes the problem with `lodash-es`, but it doesn't solve the same problem with our SDK.

davidyuk:
Let's use `@uvue/vue-cli-plugin-ssr` package (#908) because it is better maintained.

PR #911, opened 2020-12-07
Replace document.title with vue-meta

### LINKED ISSUE #910: Use vue-meta to manage pages titles
https://www.npmjs.com/package/vue-meta

This refactoring is needed for SSR. To don't refer to `document` explicitly in the router.

https://github.com/aeternity/superhero-ui/blob/3b07c6deb19a60eddcca173baad7365936a51d0f/src/router.js#L194

shapkarin:
closes #910

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -110,6 +110,14 @@ export default {
       await this.reloadUserData();
     },
   },
+  metaInfo() {
+    return {
+      // default title
+      title: 'Superhero.com',
+      // will be injected into page meta
+      titleTemplate: `${this.$route.meta.title} | Superhero.com`,
+    };
+  },
 };
 </script>
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -4346,10 +4346,10 @@
       "integrity": "sha512-e5pEa2kBnBOgR4Y/p20pskXI74UEz7de8ZGVo58asOtvSVG5YAbJeELPZxOmt+Bnz3rX753YKhfIn4X4l1PPRQ=="
     },
     "aeternity-fungible-token": {
-      "version": "github:mradkov/aeternity-fungible-token#f0632e890bbd2683f1fd757a979fe167d6d5654d",
+      "version": "github:mradkov/aeternity-fungible-token#edd903221db2516ebf9a2d6584db1dab261a06b8",
       "from": "github:mradkov/aeternity-fungible-token",
       "requires": {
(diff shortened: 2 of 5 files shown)


src/App.vue
```diff
@@ -110,6 +110,14 @@ export default {
       await this.reloadUserData();
     },
   },
+  metaInfo() {
+    return {
+      // default title
+      title: 'Superhero.com',
```
davidyuk, line 116: This is not working this way

src/App.vue
```diff
@@ -110,6 +110,14 @@ export default {
       await this.reloadUserData();
     },
   },
+  metaInfo() {
+    return {
+      // default title
```
davidyuk, line 115: Let's leave it for the documentation of `vue-meta`

src/App.vue
```diff
@@ -110,6 +110,14 @@ export default {
...
+      title: 'Superhero.com',
+      // will be injected into page meta
+      titleTemplate: `${this.$route.meta.title} | Superhero.com`,
```
davidyuk, line 118:
```suggestion
      titleTemplate: `${this.$route.meta.title} - Superhero.com`,
```
be consistent with the previous implementation, I don't want to change the title format now

davidyuk (commented):
Why you didn't add meta for each page? See how it is powerful using this way: a41e1ce888e92919885e8c64980a0d0f1fb77ef8
Proposing to merge #915 instead

src/App.vue
shapkarin: It default fallback for the case when route don't have any meta title

src/App.vue
shapkarin: ok

shapkarin:
> Why you didn't add meta for each page? See how it is powerful using this way: [a41e1ce](https://github.com/aeternity/superhero-ui/commit/a41e1ce888e92919885e8c64980a0d0f1fb77ef8)
> Proposing to merge #915 instead

Not sure why you decide to place the title to the component but OK 
My implementation does pretty the same thing – show titles. But they defined at routes as it was before.

shapkarin:
@davidyuk you think that it makes easier to keep all about page in the page component itself, right?
@mradkov please, share your thoughts too maybe @davidyuk point is right and it's better to place titles to the each component.

src/App.vue
shapkarin: right

PR #922, opened 2020-12-14
Change path of user page, add fallback redirect

### LINKED ISSUE #896: Rename /user-profile/{address} to /users/{address}
- "profile" adds unnecessary complexity
- the plural form is kind of related to https://github.com/aeternity/ae_mdw/issues/109

davidyuk:
closes #896 

https://router.vuejs.org/guide/essentials/redirect-and-alias.html
https://router.vuejs.org/api/#router-resolve

> This is a good proposition but we should have in mind the existing profile tips which point to /user-profile/...

I think this is fine to add a fallback redirect

### DIFF
--- a/src/components/TipInput.vue
+++ b/src/components/TipInput.vue
@@ -129,7 +129,8 @@ export default {
         return `https://superhero.com/tip/${this.comment.tipId}/comment/${this.comment.id}`;
       }
       if (this.userAddress) {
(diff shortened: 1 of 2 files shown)


mradkov (approved):
Looks good to me.

stanislav-slavov (approved):

PR #927, opened 2020-12-14
Comment icon doesnt redirect to tip url in a new tab

stanislav-slavov:


src/components/tipRecords/TipRecord.vue
```diff
@@ -39,6 +39,7 @@
           <div
             class="tip__comments"
             :class="[{ 'tip__comments--hascomments': tip.commentCount }]"
+            @click.stop="$route.params.tipId === tip.id ? null : goToTip()"
```
davidyuk, line 42: - I think to refer by route name is clearer
- can we write it without binding the same handler twice in the case of tip list?

davidyuk (commented):
@onvisions  Is it correct from UX perspective to have background clickable, but the icon above not?
@stanislav-slavov Shouldn't we change the cursor above the comments icon to default instead of pointer?

davidyuk (approved):
Looks acceptable

mradkov (approved):

PR #938, opened 2020-12-17
Used grid to simplify TipMedia component

### LINKED ISSUE #931: This markup is overcomplicated, can't the same be done using flexbox or grid?
This markup is overcomplicated, can't the same be done using flexbox or grid?

_Originally posted by @davidyuk in https://github.com/aeternity/superhero-ui/pull/889#discussion_r534003310_

AtanasKrondev:
Fixes #931

Grid is used to achieve the same look, but the markup  is shorter and more readable now.

### DIFF
--- a/src/components/tipRecords/TipMedia.vue
+++ b/src/components/tipRecords/TipMedia.vue
@@ -7,51 +7,21 @@
     />
     <div
       class="media-container"
+      :class="`template${Math.min(media.length, 4)}`"
       @click.stop="openGallery = true"
     >
       <img
-        v-if="media.length === 1"
-        :src="media[0]"
-        class="full"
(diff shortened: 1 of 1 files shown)


src/components/tipRecords/TipMedia.vue
```diff
@@ -7,51 +7,21 @@
     />
     <div
       class="media-container"
+      :class="`template${media.slice(0,4).length}`"
```
davidyuk, line 10: Also you can do:
```suggestion
      :class="`template${Math.min(media.length, 4)}`"
```

davidyuk (approved):

mradkov (approved):

PR #950, opened 2020-12-25
Refactoring

CedrikNikita:


src/components/AeButton.vue
```diff
@@ -1,8 +1,7 @@
 <template>
   <button
     :disabled="disabled"
-    class="btn btn-primary ae-button"
-    :class="{ green }"
+    :class="['btn btn-primary ae-button', { green }]"
```
davidyuk, line 4: we need to vote about this! I see that this way we won't write `class` twice and save a line of code, but `}]` seems more difficult to read than `}`. Worth mentioning that vue works with both of them and linter doesn't complain. I think the best is to ignore such inconsistences, write as you want, and to don't spend time rewritting that explicitly in several places (as you did😛).

davidyuk (approved):
I like all changes except `class => :class`, but I'm fine to merge that either

PR #952, opened 2020-12-29
Display largest number of FT tip amount when there is no aeternity value

stanislav-slavov:


src/components/FiatValue.vue
```diff
@@ -34,7 +33,10 @@ export default {
     },
     fiatValue() {
       if (!this.rate) return null;
-      if (this.token && !this.tokenData) return null;
+      if (this.token && !this.tokenData) {
+        return new BigNumber(0).toNumber()
+          .toLocaleString('en-US', { style: 'currency', currency: this.showCurrency });
```
davidyuk, line 38:
```suggestion
        return (0).toLocaleString('en-US', { style: 'currency', currency: this.showCurrency });
```
BigNumber is not needed to do this 🤔

src/components/FiatValue.vue
```diff
@@ -1,6 +1,5 @@
 <template>
   <span
-    v-if="fiatValue"
```
davidyuk, line 3: Won't it show `~ null` in the case [if `this.rate` is falsy](https://github.com/aeternity/superhero-ui/blob/e7975882f7cf87748520e7d7a84ebf3ba2137d47/src/components/FiatValue.vue#L35)?

src/components/TipInput.vue
```diff
@@ -121,6 +123,25 @@ export default {
         };
       },
     }),
+    largestFtTipAmount() {
+      return this.tipUrlStats.tokenTotalAmount.length
+        ? [...this.tipUrlStats.tokenTotalAmount].sort((a, b) => (
+          shiftDecimalPlaces(new BigNumber(b.amount), -this.tokenInfo[b.token].decimals)
```
davidyuk, line 129: `shiftDecimalPlaces` [already](https://github.com/aeternity/superhero-ui/blob/32076e07b0ac1ad249be984995546e2505fc7890/src/utils/index.js#L7) does the conversion to BigNumber, use one of:
```suggestion
          new BigNumber(b.amount).shiftedBy(-this.tokenInfo[b.token].decimals)
```
or
```suggestion
          shiftDecimalPlaces(b.amount, -this.tokenInfo[b.token].decimals)
```

src/components/TipInput.vue
```diff
@@ -121,6 +123,25 @@ export default {
         };
       },
     }),
+    largestFtTipAmount() {
+      return this.tipUrlStats.tokenTotalAmount.length
+        ? [...this.tipUrlStats.tokenTotalAmount].sort((a, b) => (
```
davidyuk, line 128: Actually, you don't need to sort array to get the max value, I would use `reduce` instead 🤪
```suggestion
        ? this.tipUrlStats.tokenTotalAmount.reduce(
          (a, b) => a.amount > b.amount ? a : b, // use the right comparison from the below
          this.tipUrlStats.tokenTotalAmount[0],
        )
```

src/components/TipInput.vue
```diff
@@ -10,8 +10,9 @@
       <img :src="iconTip">
       <AeAmountFiat
         v-if="!userAddress
-          && (!tipUrlStats.tokenTotalAmount.length || +tipUrlStats.totalAmountAe !== 0)"
-        :amount="tipUrlStats.totalAmountAe"
+          && (tipUrlStats.tokenTotalAmount.length || +tipUrlStats.totalAmountAe !== 0)"
```
davidyuk, line 13: you have already implemented logic to calculate value/token below, why don't reuse it?
```
v-if="!userAddress && tipAmount.value"
```

mradkov (commented):
Looks good to me.

mradkov (approved):

src/components/TipInput.vue
```diff
@@ -121,6 +121,25 @@ export default {
         };
       },
     }),
+    largestFtTipAmount() {
+      return this.tipUrlStats.tokenTotalAmount.length
+        ? this.tipUrlStats.tokenTotalAmount.reduce(
+          (a, b) => (a.amount > b.amount ? a : b),
```
davidyuk, line 127: @stanislav-slavov Shouldn't `a.amount` and `b.amount` be shifted by the decimals value of the corresponding token as it was before?

PR #897, opened 2020-11-26
New branding and identity

davidyuk:
based on #893 

changes made:
- remove unused fonts and icons
- reverted unnecessary changes in package-lock
- squashed the overlapping changes
- some changes extracted into separate commits

you can compare this branch with #893 using `git diff origin/feature/new-branding origin/new-branding`

PR #975, opened 2021-01-18
Warn user if tipping `http` URL

### LINKED ISSUE #953: Warn if users try to tip `http` URL

imollov:
closes #953

url_status_not_secure


src/assets/notSecureUrl.svg
```diff
@@ -0,0 +1,12 @@
+<svg width="24" height="24" viewBox="2 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
mradkov, line 1:
```suggestion
<svg viewBox="2 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```

mradkov: Check [this](https://github.com/aeternity/superhero-ui/pull/897#discussion_r556475843) discussion.

src/utils/index.js
```diff
@@ -61,6 +61,7 @@ export const urlStatus = (tipUrl, verifiedUrls, blacklistedUrls) => {
   const url = twitterProfile || tipUrl;
 
   if (blacklistedUrls.some((u) => url.includes(u))) return 'blacklisted';
+  if (url.startsWith('http') && !url.startsWith('https:')) return 'not-secure';
```
AtanasKrondev, line 64:
```suggestion
  if (!url.startsWith('https:')) return 'not-secure';
```

Shouldn't be enough? 
Looks good in general

AtanasKrondev (approved):

src/utils/index.js
imollov: It avoids showing the `not-secure` status prematurely to the user when the URL is manually typed.

Issue #963, opened 2021-01-07
The animation of the new loading indicator doesn't behave as expected if .svg file is imported


stanislav-slavov:
https://github.com/aeternity/superhero-ui/blob/84ccfa809bab31dd822323b3111787ff85ef014a/src/components/Loader.vue#L30

The svg file containing the loader and animation should not be inlined but extracted into a separate file.
The aforementioned link points to a "transform" property which behaves differently if the svg file is extracted from the vue component and included.

The scope of this issue is to fix the animation while keeping the vue component free from the inline < svg >.

PR #981, opened 2021-01-19
Import SVG without breaking the animation

imollov:
fixes: #963


mradkov (approved):
Looks good and it works properly now. Thanks!

PR #951, opened 2020-12-28
Feature/aeternity to store (refactoring for the SSR)

shapkarin:
Refactoring the project. Move all possible Aeternity utils to the store by the Vuex module.

Hello @Liubov-crypto 
Please test:
- tippting (with extension and web wallet)
- retip (with extension and web wallet)
- comment (with extension and web wallet)
- it is shows right account balance (with extension)

--------------

close #909

PR #985, opened 2021-01-21
Refactor aeternity module

CedrikNikita:
introduced in #951 
fixes:
- getting  token balance
- using iframe wallet
- right section being toggleable
- editing profile
- sending post report


src/store/modules/aeternity.js
```diff
@@ -13,6 +13,7 @@ import { BigNumber } from 'bignumber.js';
 import { IS_MOBILE_DEVICE } from '../../utils';
 
 export default {
+  namespaced: true,
```
shapkarin, line 16: Agree, we should to use it because it's more useful than keep modules without specific namespace.

Liubov-crypto:
* getting token balance - pass
* using iframe wallet - pass
* right section being toggleable - pass
* editing profile - pass
* sending post report - pass

Small design bugs found in FF:

white stripe in FF iframe
white stripe in FF

PR #997, opened 2021-01-26
WordBazaar - Get AE page, How it works page, Design Adjustments

AtanasKrondev:
- Get AE page with links to platforms to top-up AE tokens

- How it works page

- Adjusting some of the colours to the new design


thepiwo (approved):
lgtm

PR #1002, opened 2021-01-27
Bonding Curve Graph

### LINKED ISSUE #980: Bonding Curve Visual

venimus:
close #980


src/components/BondingCurve.vue
```diff
@@ -0,0 +1,158 @@
...
+    },
+  },
+  mounted() {
+    this.renderChart(this.chartData, this.chartOptions);
+  },
+  computed: {
+    chartData() {
+      const sellPrice = this.data.sellPrice / 1000000000000000000;
+      const buyPrice = this.data.buyPrice / 1000000000000000000;
+      const supply = this.data.totalSupply / 1000000000000000000;
```
thepiwo, line 20: these values can overrun js number precision, we should use bignumber for such calculations

thepiwo, line 20: decimals should probably not be hardcoded like this

src/components/BondingCurve.vue
```diff
@@ -0,0 +1,158 @@
...
+              { x: supply, y: 0 },
+            ],
+            borderColor: '#1161fe',
+            borderDash,
+          },
+          {
+            order: 4,
+            data: [
+              { x: 0, y: 0 },
+              { x: 3 * supply, y: 3 * sellPrice },
```
thepiwo, line 56: where does 3 come from? what if a different bonding curve contract is used?

src/views/WordDetail.vue
```diff
@@ -67,6 +67,84 @@
...
+                      :amount="data.spread"
+                      aettos
+                    />
+                  </div>
+                </div>
+                <div class="legend-item">
+                  <h3>{{ $t('views.WordDetail.LegendReserve') }}</h3>
+                  <div>
+                    <AeAmount
+                      :amount="data.sellPrice * data.totalSupply / 2000000000000000000"
```
thepiwo, line 140: whats that devision by? no magic numbers please

thepiwo (commented):
- chart is not changing if supply changed
- sell line is not shown if supply < 1
- data seems incorrect or visualization is missleading, e.g. when I want to sell 1.5, I should receive 1.5 in this example, but the graph shows 1.25 (sell price for 1) same for buying values that are not 1
- also that graph lets the user know the price rises to ~4.3 if supply is 2, actually price will be 3.5 if supply is 2
image

src/components/BondingCurve.vue
thepiwo: no magic numbers please

src/views/WordDetail.vue
venimus: Actually not sure how "Reserve" should be calculated, I assumed it is the area,

src/components/BondingCurve.vue
venimus: we would need to get some prediction for the future to display, that is hinted in the design (~3 times the supply). Since it is a linear curve that starts from 0, it is a simple multiplication.
Is there a way to get appropriate data?

mradkov: You can fetch prices straight from the contract. Given a total supply and amount for buy/sell you can get the price.

src/views/WordDetail.vue
mradkov: The reserve is everything below the sell curve.

venimus:
The chart shows the Bonding curve, so values are not supposed to match the "Current Buy/Sell" prices but rather the calculation parameters based on the current supply and Spread is not the historically "accumulated spread".

package.json
```diff
@@ -24,6 +24,7 @@
     "axios": "^0.21.1",
     "bignumber.js": "^9.0.1",
     "bootstrap": "^4.5.3",
+    "chart.js": "^2.9.4",
```
mradkov, line 27: Is this used anywhere?

venimus: vue-chart.js requires it (apparently not added to internal deps)
https://vue-chartjs.org/guide/#installation

thepiwo (approved):
looks good to me

thepiwo:
@mradkov feel free to rebase and merge

AtanasKrondev (approved):

mradkov (approved):

PR #993, opened 2021-01-22
Refactoring: replace utils/backend with Vuex modules/backend

shapkarin:
Currently the propose of this PR is to achieve the SSR functionality for the rest of the app – at least to make `Tip page` and `Profile page` sharable with social networks.
This PR is the second part of that PRs: https://github.com/aeternity/superhero-ui/pull/951 https://github.com/aeternity/superhero-ui/pull/985

Issue #909, opened 2020-12-06
Don't import store explicitly


davidyuk:
https://github.com/aeternity/superhero-ui/blob/78572cc2834215b08900226e92790cc230dffb6f/src/utils/aeternity.js#L11

https://github.com/aeternity/superhero-ui/blob/ab19bbfc47534148b432e252e13eb5f39b9e4ae7/src/utils/backend.js#L2

Probably the best would be to rewrite backend.js and aeternity.js as Vuex modules. This refactoring is needed for SSR.

davidyuk:
the related problem is a bit explained here: https://github.com/aeternity/superhero-ui/pull/901/files#r539518456

shapkarin:
> https://github.com/aeternity/superhero-ui/pull/901/files#r539518456

thanks a lot

shapkarin:
Compare with develop: https://github.com/aeternity/superhero-ui/compare/feature/aeternity_to_store

TODO: move Aeternity parts to the module https://vuex.vuejs.org/guide/modules.html

davidyuk:
Well, the store is still imported in both `utils/aeternity.js` and `utils/backend.js`

shapkarin:
> Well, the store is still imported in both `utils/aeternity.js` and `utils/backend.js`

thanks, didn't notice that. And, right, `utils/backend.js` also needs to be refactored.

shapkarin:
@davidyuk think to add store `getters` later for the `modules/aeternity.js` and `modules/backend.js`.
looks like it can be useful.

shapkarin:
@davidyuk @mradkov @CedrikNikita do you think `wrapTry` and `backendFetch` are should be an the store `actions`? 
I don't think so. 

What do you think? I already made them to be the actions but I think to revert that later.

https://github.com/aeternity/superhero-ui/blob/f484d2b4b4808a120218d2ebf564a7b59f67e0c9/src/utils/backend.js#L4-L22

https://github.com/aeternity/superhero-ui/blob/f484d2b4b4808a120218d2ebf564a7b59f67e0c9/src/utils/backend.js#L24-L26

shapkarin:
@davidyuk @mradkov @CedrikNikita I also suggest to avoid Vue `mixins` as much as possible and use Vuex features instead.

davidyuk:
fixed in #1115

PR #1003, opened 2021-01-27
Refact: don't import store explicitly at the utils/backend.js

### LINKED ISSUE #707: Don't omit exceptions thrown from backend fetch
https://github.com/aeternity/superhero-ui/blob/a60c0d4bad5fcf85a22d47753d5032d12e1d5c3d/src/utils/backend.js#L5

shapkarin:
issue: #909 
ref: https://github.com/aeternity/superhero-ui/pull/993

upd:
closes #707
closes aeternity/superhero-wallet#548

### DIFF
--- a/src/utils/backend.js
+++ b/src/utils/backend.js
@@ -9,10 +9,8 @@ const wrapTry = async (promise) => {
         return null;
       }
       if (!res.ok) EventBus.$emit('maintenance');
+      EventBus.$emit('backToFeed');
       return res.json();
-    }).catch(() => {
-      EventBus.$emit('maintenance');
-      return null;
     });
   } catch (err) {
     EventBus.$emit('maintenance');
--- a/src/utils/backend.js
+++ b/src/utils/backend.js
@@ -1,22 +1,21 @@
 // eslint-disable-next-line import/no-cycle
-import store from '../store';
+import { EventBus } from './eventBus';
 
 const wrapTry = async (promise) => {
   try {
     return promise.then((res) => {
       if (!res) {
-        store.commit('setBackendStatus', false);
+        EventBus.$emit('maintenance');
         return null;
       }
-      store.commit('setBackendStatus', true);
-      if (!res.ok) throw new Error(`Request failed with ${res.status}`);
+      if (!res.ok) EventBus.$emit('maintenance');
       return res.json();
-    }).catch((error) => {
-      console.error(error);
+    }).catch(() => {
+      EventBus.$emit('maintenance');
       return null;
     });
   } catch (err) {
-    store.commit('setBackendStatus', false);
+    EventBus.$emit('maintenance');
     return null;
   }
 };
--- a/src/App.vue
+++ b/src/App.vue
@@ -52,6 +52,8 @@ export default {
       this.initWallet(),
       this.reloadData(),
     ]);
+
+    EventBus.$on('maintenance', () => this.$router.push({ name: 'maintenance' }));
   },
   methods: {
     ...mapMutations([
--- a/src/App.vue
+++ b/src/App.vue
@@ -46,7 +46,6 @@ export default {
       this.reloadData();
     });
     setInterval(() => this.reloadData(), 120 * 1000);
-
     await this.$store.dispatch('aeternity/initSdk');
     await Promise.all([
       this.initWallet(),
@@ -54,6 +53,11 @@ export default {
     ]);
 
     EventBus.$on('maintenance', () => this.$router.push({ name: 'maintenance' }));
+    EventBus.$on('backToFeed', () => {
+      if (this.$route.name === 'maintenance') {
+        this.$router.push({ name: 'feed' });
+      }
+    });
   },
   methods: {
     ...mapMutations([


shapkarin:
@Liubov-crypto please test that:
- open the app
- switch off the internet to produce the request error
- check if the `Maintenance page` is showed

shapkarin:
@mradkov basically all is fine and it's enough to integrate the SSR because the main goal of that task is to not import store explicitly at the `utils/backend.js`.
I already tested [that](https://github.com/aeternity/superhero-ui/pull/1003#issuecomment-768492597) but we can wait for @Liubov-crypto to make sure.

My previous PR https://github.com/aeternity/superhero-ui/pull/993 is marked as draft but later we can continue with more deeply approach that present at that PR or connect SDK and swagger to replace the `utils/backend.js` with SDK methods that was grabbed from the swagger file. Me or @davidyuk can explain that, if you are not aware about the last.

shapkarin:
thanks to @davidyuk 
it also can use `window.addEventListener("unhandledrejection", function (event) {` instead to catch the manually defined `throw new error` from the fetch.

shapkarin:
> thanks to @davidyuk
> it can also use `window.addEventListener("unhandledrejection", function (event) {` instead to catch the manually defined `throw new error` from the fetch.

refs: 
https://github.com/aeternity/superhero-ui/issues/707
https://github.com/aeternity/superhero-wallet/issues/548

shapkarin:
there is some blockers with `unhandledrejection` so I will leave it with `eventBus`

src/utils/backend.js
```diff
@@ -1,22 +1,24 @@
...
   try {
     return promise.then((res) => {
       if (!res) {
-        store.commit('setBackendStatus', false);
+        EventBus.$emit('maintenance');
         return null;
       }
-      store.commit('setBackendStatus', true);
-      if (!res.ok) throw new Error(`Request failed with ${res.status}`);
+      // store.commit('setBackendStatus', true);
```
mradkov, line 11: When does the maintenance page return back to the main feed, if you comment this out?

shapkarin: I will remove that comment. It works with eventBus now:
next line is `EventBus.$emit('maintenance');`
and at the App.vue i have a handler:
```js
EventBus.$on('maintenance', () => this.$router.push({ name: 'maintenance' }));
```

shapkarin: so basically it's:
https://github.com/aeternity/superhero-ui/blob/f55bc1918c0e51011772c999b653dae9bef68ae7/src/utils/backend.js#L8
and
https://github.com/aeternity/superhero-ui/blob/f55bc1918c0e51011772c999b653dae9bef68ae7/src/App.vue#L56

Liubov-crypto:
> @Liubov-crypto please test that:
> 
> * open the app
> * switch off the internet to produce the request error
> * check if the `Maintenance page` is showed

Followed steps described above. Got this in Chrome (the picture of the blue diamond was not loaded):

ууу

Also, when the Internet appears there is no automatic redirection to the main page, the user remains on the maintenance page:
rr

FF - maintenance text and the picture of the blue diamond was not loaded:
мозіла

Brave - maintenance text and the picture of the blue diamond was not loaded:
брейв1

shapkarin:
@Liubov-crypto thanks, need to investigate what's going on in Brave

shapkarin:
currently I got that page on error, but from time to time it not appears. need to investigate in any case...
Снимок экрана 2021-01-28 в 11 56 19

shapkarin:
@Liubov-crypto try to reproduce it that way in Brave:
- open profile page
- switch off the internet
- open main page (tips list page)

I got the `maintenance page` in that way. Not sure why it's not fires all the time. Looks like Brave has other rules for fire an error on `internet disconnected` but the same for the other errors. Maybe that's related to the manually switch off the internet.

Liubov-crypto:
> @Liubov-crypto try to reproduce it that way in Brave:
> 
> * open profile page
> * switch off the internet
> * open main page (tips list page)
> 
> I got the `maintenance page` in that way. Not sure why it's not fires all the time. Looks like Brave has other rules for fire an error on `internet disconnected` but the same for the other errors. Maybe that's related to the manually switch off the internet.

Works in this way. But after connecting to the Internet, the user remains on the maintenance page. No automatic redirection to the main page.

shapkarin:
> > @Liubov-crypto try to reproduce it that way in Brave:
> > 
> > * open profile page
> > * switch off the internet
> > * open main page (tips list page)
> > 
> > I got the `maintenance page` in that way. Not sure why it's not fires all the time. Looks like Brave has other rules for fire an error on `internet disconnected` but the same for the other errors. Maybe that's related to the manually switch off the internet.
> 
> Works in this way. But after connecting to the Internet, the user remains on the maintenance page. No automatic redirection to the main page.

Yes, the logic was changed. @davidyuk suggest to do that. But I can try to forward it to the main on connect.

shapkarin:
@Liubov-crypto I guess now it can be forwarded to main page after the maintenance

Liubov-crypto:
> @Liubov-crypto I guess now it can be forwarded to main page after the maintenance

now redirects to the main page, but periodically redirects to https://feature-refact-backend.aggregator.z52da5wt.xyz/maintenance 

also, I can't stay on my profile page, redirection to the main page after a few seconds

src/utils/backend.js
mradkov: And how the user gets redirected back to the feed when the maintenance is finished?

shapkarin:
@Liubov-crypto thanks. It's a bug

src/utils/backend.js
shapkarin: > When does the maintenance page return back to the main feed, if you comment this out?

@mradkov right, need to solve that

shapkarin: @mradkov not sure that it's possible now

shapkarin: @mradkov I guess I solved that. Need to recheck.

shapkarin:
@Liubov-crypto after my last commit it should redirects to the feed page (the main page). Please, check.

src/utils/backend.js
shapkarin: @mradkov I don't like all this PR at all but I have no idea how to easily achieve the SSR (remove import store explicitly) 
This should work as a temporary solution. It also redirects to the feed page with the last commit https://github.com/aeternity/superhero-ui/pull/1003/commits/6e1db153b3e122c5234b1af198d47040edfa4687

Liubov-crypto:
> @Liubov-crypto after my last commit it should redirects to the feed page (the main page). Please, check.

Followed these steps: 
- open profile page
- switch off the internet
- open main page (tips list page)
- turn on the internet
- after a while the app should show the feed list

Got the issue that the main page redirected to the maintenance page after a while. 
I recorded a video, the same behaviour in FF and Brave: 


https://user-images.githubusercontent.com/69896204/106155007-e0ce6000-6188-11eb-8540-6243c3f0e404.mov

shapkarin:
@Liubov-crypto some loop redirects... after a while it's fine. I will investigate.

shapkarin:
@Liubov-crypto pushed the commit, I checked, it's fine with Brave now. Cam you confirm?

Liubov-crypto:
> @Liubov-crypto pushed the commit, I checked, it's fine with Brave now. Can you confirm?

Followed these steps:

open profile page
switch off the internet
open main page (tips list page)
turn on the internet
after a while the app should show the feed list

Now everything looks good, no redirection to maintenance. 
btw, there is some delay in loading the feed in FF.

mradkov (dismissed):
As a way to get it working, this look acceptable. Though we need to get rid of all EventBus usages in the future - logged an issue for this [here](https://github.com/aeternity/superhero-ui/issues/1006).

Liubov-crypto:
Btw, there is no maintenance banner in Safari, just this: 
саф

After turning on the Internet, you need to refresh the page so that the feed appears

PR #1005, opened 2021-01-28
WordBazaar - Switch token info on right side, Buy/Sell price in listing

### LINKED ISSUE #1004: WordBazaar - switch right side wallet to word detail page if user has balance of that token

AtanasKrondev:
Closes #1004 
-Switch token info on right side

- Reduced wording related warnings
- Buy/Sell price in listing

### DIFF
--- a/src/components/WordListing.vue
+++ b/src/components/WordListing.vue
@@ -21,12 +21,13 @@
       <div class="word-listing-column">
         <span
           class="heading-text"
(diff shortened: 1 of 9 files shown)


thepiwo (commented):
maybe we should switch back to AE if navigating elsewhere, or show two lines of balances as we discussed then this is not an issue

thepiwo (approved):
looks good!

PR #1008, opened 2021-01-29
WordBazaar - Fix word-dropdown, not triggering, message not changing on sell

AtanasKrondev:
Compare 'word' with wordRegistry instead with tokenInfo


thepiwo (approved):
looks good

PR #1024, opened 2021-02-22
Middleware access with swagger

### LINKED ISSUE #1017: Use swagger file to access middleware endpoints
https://github.com/aeternity/superhero-ui/blob/62857bff23ab868de949325c3e627c27043c7437/src/utils/middleware.js#L3-L9

See how it works in the wallet as an example:

https://github.com/aeternity/superhero-wallet/blob/92a8f0053c2bfc3fcb8506c917f33103d960412f/src/lib/wallet.js#L12-L68

cc: @thepiwo @AtanasKrondev

AtanasKrondev:
Fixes #1017

### DIFF
--- a/src/store/index.js
+++ b/src/store/index.js
@@ -1,6 +1,8 @@
 import Vue from 'vue';
 import Vuex from 'vuex';
 import BigNumber from 'bignumber.js';
+import Swagger from '@aeternity/aepp-sdk/es/utils/swagger';
+import { camelCase } from 'lodash-es';
 import mutations from './mutations';
 import getters from './getters';
 import persistState from './plugins/persistState';
@@ -10,7 +12,6 @@ import backend from './modules/backend';
 import aeternity from './modules/aeternity';
 // eslint-disable-next-line import/no-cycle
 import Backend from '../utils/backend';
-import Middleware from '../utils/middleware';
 
 Vue.use(Vuex);
 
@@ -34,6 +35,7 @@ export default new Vuex.Store({
     isHiddenContent: true,
     isBackendLive: true,
     cookiesConsent: {},
+    middleware: null,
   },
   mutations,
   actions: {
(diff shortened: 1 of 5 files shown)


src/utils/middleware.js
```diff
@@ -1,9 +1,43 @@
...
 );
 
-export default class Middleware {
-  static getAex9Balance = async (token, account) => backendFetch(`aex9/balance/${token}/${account}?top=true`);
+export default async function initMiddleware() {
+  const swag = await middlewareFetch('swagger/swagger.json');
+
+  swag.paths = {
+    ...swag.paths,
+    'aex9/balance/{token}/{account}?top=true': {
```
davidyuk, line 14: Have you created an issue in middleware asking to explain this method in the swagger file?

src/utils/middleware.js
```diff
@@ -1,9 +1,43 @@
...
 import { wrapTry } from './index';
 
-const backendFetch = (path, ...args) => wrapTry(
+const middlewareFetch = (path, ...args) => wrapTry(
   fetch(`${process.env.VUE_APP_MIDDLEWARE_URL}/${path}`, ...args).catch((err) => console.error(err)),
 );
 
-export default class Middleware {
-  static getAex9Balance = async (token, account) => backendFetch(`aex9/balance/${token}/${account}?top=true`);
+export default async function initMiddleware() {
```
davidyuk, line 9: I think you can put this whole function to `initMiddleware` action because it will be used only there.

src/utils/middleware.js
```diff
@@ -1,9 +1,43 @@
+import Swagger from '@aeternity/aepp-sdk/es/utils/swagger';
+import { camelCase } from 'lodash-es';
 import { wrapTry } from './index';
 
-const backendFetch = (path, ...args) => wrapTry(
+const middlewareFetch = (path, ...args) => wrapTry(
```
davidyuk, line 5: I don't think that `wrapTry` wrapper is needed here, as we don't do that for requesting a swagger file of the node.

src/store/index.js
```diff
@@ -68,6 +70,11 @@ export default new Vuex.Store({
         }
       }));
     },
+    async initMiddleware({ commit }) {
+      const middleware = await initMiddleware();
+      console.log(middleware);
```
davidyuk, line 75: Ensure that this is removed before merging.

davidyuk (approved):

src/utils/middleware.js
thepiwo: the mdw team is on it

mradkov (approved):

PR #1027, opened 2021-03-03
feat: global error handler, ErrorReportModal

shapkarin:
This PR includes:
- throw as much errors as possible instead of console logs
- error popup for all unhandled exceptions with it's trace stack that showed one by one
  - if one is closed and there is one more error the new one appears

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "connect": "^3.7.0",
     "core-js": "^3.15.1",
     "date-fns": "^2.22.1",
+    "detect-browser": "^5.2.0",
     "fontsource-ibm-plex-sans": "^3.1.5",
     "i18n": "^0.13.3",
     "is-fqdn": "^2.0.1",
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "lru-cache": "^6.0.0",
     "node-fetch": "^2.6.1",
     "normalize.css": "^8.0.1",
+    "serialize-error": "^8.1.0",
     "serve-static": "^1.14.1",
     "sophia-bonding-curve": "github:aeternity/BondingCurve#1.0.0-alpha.2",
     "soundcloud-widget": "^0.2.1",
--- a/src/App.vue
+++ b/src/App.vue
@@ -101,7 +101,11 @@ export default {
         console.log('found wallet');
         this.useSdkWallet();
         this.setAddress(address);
-        this.$store.dispatch('updateCookiesConsent');
+        try {
+          await this.$store.dispatch('updateCookiesConsent');
+        } catch (error) {
+          if (error.message !== 'Operation rejected by user') throw error;
+        }
       } else {
         this.setAddress(address);
       }
--- a/src/components/UserInfo.vue
+++ b/src/components/UserInfo.vue
@@ -175,7 +175,7 @@
             :key="scope"
             class="cookies-button"
             :class="{ active: cookiesConsent[scope] }"
-            @click="setCookies({ scope, status: !cookiesConsent[scope] })"
+            @click="toggleCookiesConsent(scope)"
           >
             {{ scope }}
           </ButtonPlain>
@@ -381,6 +381,16 @@ export default {
       this.profile = await Backend.getProfile(this.address);
       this.profile.biography = this.profile.biography || '';
       if (this.isOwn) this.$store.commit('setUserProfile', this.profile);
+    },
+    async toggleCookiesConsent(scope) {
+      try {
+        await this.$store.dispatch(
+          'backend/setCookies',
+          { scope, status: !this.cookiesConsent[scope] },
+        );
+      } catch (error) {
+        if (error.message !== 'Operation rejected by user') throw error;
+      }
     },
   },
 };
--- a/src/store/plugins/errorHandler.js
+++ b/src/store/plugins/errorHandler.js
@@ -1,0 +1,54 @@
+import { serializeError } from 'serialize-error';
+import Vue from 'vue';
+
+export default ({ dispatch }) => {
+  const reportsToSend = [];
+  let showingErrorReportModal = false;
+
+  const handleError = async (error) => {
+    reportsToSend.push({
+      appName: 'superhero-ui',
+      appVersion: process.env.npm_package_version,
+      appRevision: process.env.COMMIT_HASH,
+      userAgent: window.navigator.userAgent,
+      location: window.location.href,
+      timestamp: Date.now(),
+      error,
+    });
+    if (showingErrorReportModal) return;
+    showingErrorReportModal = true;
+    let isReportSubmitted = false;
+    while (reportsToSend.length) {
+      // eslint-disable-next-line no-await-in-loop
+      isReportSubmitted = await dispatch('modals/open', {
+        name: 'error-report-modal',
+        report: reportsToSend.shift(),
+      }) || isReportSubmitted;
+    }
+    if (isReportSubmitted) {
+      await dispatch('modals/open', { name: 'error-report-submitted-modal' });
+    }
+    showingErrorReportModal = false;
+  };
+
+  window.addEventListener('unhandledrejection', (event) => handleError({
+    exceptionType: 'unhandled-rejection',
+    ...serializeError(event.reason),
+  }));
+  window.onerror = (message, source, line, col, error) => handleError({
+    exceptionType: 'unhandled-error',
+    message,
+    source,
+    line,
+    col,
+    ...serializeError(error),
+  });
+  Vue.config.errorHandler = async (error, vm, info) => {
+    console.error(error);
+    await handleError({
+      exceptionType: 'unhandled-vue-error',
+      info,
+      ...serializeError(error),
+    });
+  };
+};
--- a/src/store/plugins/errorHandler.js
+++ b/src/store/plugins/errorHandler.js
@@ -1,5 +1,6 @@
 import { serializeError } from 'serialize-error';
 import Vue from 'vue';
+import { detect } from 'detect-browser';
 
 export default ({ dispatch }) => {
   const reportsToSend = [];
@@ -7,13 +8,15 @@ export default ({ dispatch }) => {
 
   const handleError = async (error) => {
     reportsToSend.push({
-      appName: 'superhero-ui',
       appVersion: process.env.npm_package_version,
-      appRevision: process.env.COMMIT_HASH,
-      userAgent: window.navigator.userAgent,
-      location: window.location.href,
-      timestamp: Date.now(),
-      error,
+      browser: detect(),
+      platform: 'superhero-ui',
+      time: Date.now(),
+      error: {
+        ...error,
+        appRevision: process.env.COMMIT_HASH,
+        location: window.location.href,
+      },
     });
     if (showingErrorReportModal) return;
     showingErrorReportModal = true;
--- a/src/utils/backend.js
+++ b/src/utils/backend.js
@@ -160,4 +160,12 @@ export default class Backend {
     body: JSON.stringify(postParam),
     headers: { 'Content-Type': 'application/json' },
   });
+
+  static errorReport = (report) => {
+    backendFetch('errorreport', {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify(report),
+    });
+  };
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -8705,6 +8705,11 @@
       "resolved": "https://registry.npmjs.org/dns-equal/-/dns-equal-1.0.0.tgz",
       "integrity": "sha1-s55/HabrCnW6nBcySzR1PEfgZU0="
     },
+    "detect-browser": {
+      "version": "5.2.0",
+      "resolved": "https://registry.npmjs.org/detect-browser/-/detect-browser-5.2.0.tgz",
+      "integrity": "sha512-tr7XntDAu50BVENgQfajMLzacmSe34D+qZc4zjnniz0ZVuw/TZcLcyxHQjYpJTM36sGEkZZlYLnIM1hH7alTMA=="
+    },
     "dns-packet": {
       "version": "1.3.4",
       "resolved": "https://registry.npmjs.org/dns-packet/-/dns-packet-1.3.4.tgz",
--- a/package-lock.json
+++ b/package-lock.json
@@ -19359,6 +19359,21 @@
         }
       }
     },
+    "serialize-error": {
+      "version": "8.1.0",
+      "resolved": "https://registry.npmjs.org/serialize-error/-/serialize-error-8.1.0.tgz",
+      "integrity": "sha512-3NnuWfM6vBYoy5gZFvHiYsVbafvI9vZv/+jlIigFn4oP4zjNPK3LhcY0xSCgeb1a5L8jO71Mit9LlNoi2UfDDQ==",
+      "requires": {
+        "type-fest": "^0.20.2"
+      },
+      "dependencies": {
+        "type-fest": {
+          "version": "0.20.2",
+          "resolved": "https://registry.npmjs.org/type-fest/-/type-fest-0.20.2.tgz",
+          "integrity": "sha512-Ne+eE4r0/iWnpAxD852z3A+N0Bt5RN//NjJwRd2VFHEmrywxf5vsZlh4R6lixl6B+wz/8d+maTSAkN1FIkI3LQ=="
+        }
+      }
+    },
     "serialize-javascript": {
       "version": "4.0.0",
       "resolved": "https://registry.npmjs.org/serialize-javascript/-/serialize-javascript-4.0.0.tgz",
--- a/src/assets/iconCodeError.svg
+++ b/src/assets/iconCodeError.svg
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 25.2.1, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 viewBox="0 0 48 48" style="enable-background:new 0 0 48 48;" xml:space="preserve">
+<style type="text/css">
+	.st0{fill:none;}
+	.st1{fill-rule:evenodd;clip-rule:evenodd;fill:#FF4746;}
+	.st2{opacity:0.44;fill:#FF4746;enable-background:new;}
+</style>
+<g id="transparent">
+	<rect x="0" y="0" class="st0" width="48" height="48"/>
+</g>
+<g id="brackets">
+	<path class="st1" d="M14.337,15.404c0.731,0.688,0.766,1.839,0.078,2.57l-6.1,6.481l6.1,6.481c0.688,0.731,0.653,1.882-0.078,2.57
+		c-0.731,0.688-1.882,0.653-2.57-0.078l-7.273-7.727c-0.659-0.7-0.659-1.792,0-2.492l7.273-7.727
+		C12.455,14.751,13.606,14.716,14.337,15.404z M33.663,15.404c-0.731,0.688-0.766,1.839-0.078,2.57l6.1,6.481l-6.1,6.481
+		c-0.688,0.731-0.653,1.882,0.078,2.57c0.731,0.688,1.882,0.653,2.57-0.078l7.273-7.727c0.659-0.7,0.659-1.792,0-2.492l-7.273-7.727
+		C35.545,14.751,34.394,14.716,33.663,15.404z"/>
+</g>
+<g id="blinking">
+	<path class="st2" d="M27.765,9.951L27.765,9.951c0.97,0.26,1.546,1.257,1.286,2.227l-6.588,24.587
+		c-0.26,0.97-1.257,1.546-2.227,1.286l0,0c-0.97-0.26-1.546-1.257-1.286-2.227l6.588-24.587
+		C25.798,10.266,26.795,9.691,27.765,9.951z"/>
+		<animate attributeName="opacity"
+             values=".56;1;.56" dur="1.3s"
+             repeatCount="indefinite"/>
+</g>
+</svg>
--- a/src/assets/iconThumbUp.svg
+++ b/src/assets/iconThumbUp.svg
@@ -1,0 +1,33 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 25.2.1, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 viewBox="0 0 48 48" style="enable-background:new 0 0 48 48;" xml:space="preserve">
+<style type="text/css">
+	.st0{fill:none;}
+	.st1{opacity:0.44;fill:#00FF9D;enable-background:new;}
+	.st2{fill-rule:evenodd;clip-rule:evenodd;fill:#00FF9D;opacity:1;}
+</style>
+<g id="bound">
+	<rect x="0" y="0" class="st0" width="48" height="48"/>
+</g>
+<g id="animated1">
+	<path class="st1" d="M7.21,17h6c1.105,0,2,0.895,2,2v18c0,1.105-0.895,2-2,2h-6c-1.105,0-2-0.895-2-2V19
+		C5.21,17.895,6.105,17,7.21,17z">
+		<animate attributeName="opacity"
+             values=".44;.66;.44" dur="1.3s"
+             repeatCount="indefinite"/>
+	</path>
+</g>
+<g id="animated2">
+	<path cx="30.501" cy="24" class="st2" d="M18.21,19v18l2.305,0.768C22.962,38.584,25.525,39,28.105,39h7.187c2.86,0,5.323-2.019,5.883-4.823
+		l1.498-7.492c0.775-3.874-2.341-7.426-6.283-7.163L29.21,20l0.642-7.061C30.044,10.824,28.379,9,26.255,9l0,0
+		c-1.269,0-2.444,0.666-3.097,1.754L18.21,19z">
+		<animate attributeName="opacity"
+             values="1;.44;1" dur="1.5s"
+             repeatCount="indefinite"/>
+			 <animate attributeName="cx"
+             values="30.501;32;30.501" dur="1.1s"
+             repeatCount="indefinite"/>
+	</path>
+</g>
+</svg>
--- a/src/components/AeButton.vue
+++ b/src/components/AeButton.vue
@@ -49,6 +49,7 @@ export default {
   transition: background-color 0.3s;
   text-align: center;
   font-size: 16px;
+  cursor: pointer;
 
   &[disabled] {
     opacity: 0.4;
--- a/src/components/AlertModal.vue
+++ b/src/components/AlertModal.vue
@@ -1,25 +1,23 @@
 <template>
   <div
-    class="modal-mask"
-    @click="resolve"
+    class="alert-modal"
+    @click.self="resolve"
   >
-    <div class="modal-wrapper">
-      <div class="success-modal">
-        <img
-          v-if="!hideIcon"
-          :src="failure ? FailureIcon : SuccessIcon"
-        >
-        <h1>{{ title }}</h1>
-        <p
-          v-for="(row, index) in body instanceof Array ? body : [body]"
-          :key="index"
-        >
-          {{ row }}
-        </p>
-        <AeButton @click="resolve">
-          {{ primaryButtonText }}
-        </AeButton>
-      </div>
+    <div class="content">
+      <img
+        v-if="!hideIcon"
+        :src="failure ? FailureIcon : SuccessIcon"
+      >
+      <h1>{{ title }}</h1>
+      <p
+        v-for="(row, index) in body instanceof Array ? body : [body]"
+        :key="index"
+      >
+        {{ row }}
+      </p>
+      <AeButton @click="resolve">
+        {{ primaryButtonText }}
+      </AeButton>
     </div>
   </div>
 </template>
@@ -35,40 +33,36 @@ export default {
     title: { type: String, required: true },
     body: { type: [String, Array], required: true },
     resolve: { type: Function, required: true },
-    failure: { type: Boolean },
-    hideIcon: { type: Boolean },
+    failure: Boolean,
+    hideIcon: Boolean,
     primaryButtonText: {
       type: String,
       default() { return this.$t('done'); },
     },
   },
-  data() {
-    return {
-      SuccessIcon,
-      FailureIcon,
-    };
-  },
+  data: () => ({
+    SuccessIcon,
+    FailureIcon,
+  }),
 };
 </script>
 
 <style lang="scss" scoped>
-.modal-mask {
-  vertical-align: middle;
-  padding-top: 5rem;
+.alert-modal {
   position: fixed;
   z-index: 9998;
   top: 0;
+  right: 0;
+  bottom: 0;
   left: 0;
-  width: 100%;
-  height: 100%;
-  display: table;
+  display: flex;
   background-color: rgba(0, 0, 0, 0.5);
 
-  .success-modal {
+  .content {
     text-align: center;
     width: 25rem;
-    margin: 0 auto;
-    padding: 2.5rem 2.5rem;
+    margin: auto;
+    padding: 2.5rem;
     background-color: $actions_ribbon_background_color;
     border-radius: 0.25rem;
     border: 1px solid $card_border_color;
--- a/src/components/AlertModal.vue
+++ b/src/components/AlertModal.vue
@@ -16,12 +16,9 @@
         >
           {{ row }}
         </p>
-        <button
-          class="button"
-          @click="resolve"
-        >
+        <AeButton @click="resolve">
           {{ primaryButtonText }}
-        </button>
+        </AeButton>
       </div>
     </div>
   </div>
@@ -30,8 +27,10 @@
 <script>
 import SuccessIcon from '../assets/verifiedUrl.svg';
 import FailureIcon from '../assets/iconError.svg';
+import AeButton from './AeButton.vue';
 
 export default {
+  components: { AeButton },
   props: {
     title: { type: String, required: true },
     body: { type: [String, Array], required: true },
@@ -90,18 +89,6 @@ export default {
       color: $standard_font_color;
       font-size: 1rem;
       font-weight: 500;
-    }
-
-    .button {
-      background-color: $secondary_color;
-      border: none;
-      border-radius: 0.25rem;
-      color: $standard_font_color;
-      font-size: 0.75rem;
-      font-weight: 700;
-      justify-self: center;
-      line-height: 1.125;
-      padding: 0.65rem 1rem;
     }
   }
 }
--- a/src/components/AlertModal.vue
+++ b/src/components/AlertModal.vue
@@ -1,34 +1,34 @@
 <template>
-  <div
+  <Dialog
     class="alert-modal"
-    @click.self="resolve"
+    :title="title"
+    @close="resolve"
   >
-    <div class="content">
-      <img
-        v-if="!hideIcon"
-        :src="failure ? FailureIcon : SuccessIcon"
-      >
-      <h1>{{ title }}</h1>
-      <p
-        v-for="(row, index) in body instanceof Array ? body : [body]"
-        :key="index"
-      >
-        {{ row }}
-      </p>
-      <AeButton @click="resolve">
-        {{ primaryButtonText }}
-      </AeButton>
-    </div>
-  </div>
+    <img
+      v-if="!hideIcon"
+      slot="icon"
+      :src="failure ? FailureIcon : SuccessIcon"
+    >
+    <p
+      v-for="(row, index) in body instanceof Array ? body : [body]"
+      :key="index"
+    >
+      {{ row }}
+    </p>
+    <AeButton @click="resolve">
+      {{ primaryButtonText }}
+    </AeButton>
+  </Dialog>
 </template>
 
 <script>
 import SuccessIcon from '../assets/verifiedUrl.svg';
 import FailureIcon from '../assets/iconError.svg';
+import Dialog from './Dialog.vue';
 import AeButton from './AeButton.vue';
 
 export default {
-  components: { AeButton },
+  components: { Dialog, AeButton },
   props: {
     title: { type: String, required: true },
     body: { type: [String, Array], required: true },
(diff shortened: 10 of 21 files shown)


shapkarin:
@davidyuk I also remove all about old maintenance logic and backendLive state itself, please notice if I'm doing wrong.

shapkarin:
@davidyuk I will recheck my work and ping you

src/App.vue
```diff
@@ -42,6 +43,15 @@ export default {
     ...mapState('aeternity', ['sdk']),
   },
   async created() {
+    window.addEventListener('unhandledrejection', (error) => {
+      if (error !== 200 && this.$route.name !== 'maintenance') {
```
davidyuk, line 47: Nothing should throw errors with 200 code. Also, I would remove the maintenance page change, because currently, we have no plan to use it. Maybe we can remove it at all?

src/App.vue
```diff
@@ -109,6 +119,11 @@ export default {
       }
       await this.reloadUserData();
     },
+    show() {
+      this.$store.dispatch('modals/open', {
```
davidyuk, line 123: Don't forget to remove this debug code

src/components/ErrorModal.vue
```diff
@@ -0,0 +1,140 @@
...
+          {{ $t('sendBugReport') }}
+        </button>
+      </div>
+    </div>
+  </div>
+</template>
+
+<script>
+import { detect } from 'detect-browser';
+
```
davidyuk, line 26:
```suggestion
```

src/components/ErrorModal.vue
```diff
@@ -0,0 +1,140 @@
...
+import SuccessIcon from '../assets/verifiedUrl.svg';
+import FailureIcon from '../assets/iconError.svg';
+import { backendFetch } from '../utils/backend';
+
+export default {
+  props: {
+    failure: { type: Boolean },
```
davidyuk, line 33: Is this prop used somewhere?

src/components/ErrorModal.vue
```diff
@@ -0,0 +1,140 @@
...
+  methods: {
+    async send() {
+      const report = {
+        appVersion: process.env.npm_package_version,
+        browser: detect(),
+        error: this.error,
+        platform: 'web',
+        description: this.report,
+        time: new Date().toISOString(),
+      };
```
davidyuk, line 54: Ensure that there is stacktrace in this report. Also, add `superhero-ui` somehow to distinguish it from wallet reports.

src/components/ErrorModal.vue
```diff
@@ -0,0 +1,140 @@
...
+import SuccessIcon from '../assets/verifiedUrl.svg';
+import FailureIcon from '../assets/iconError.svg';
+import { backendFetch } from '../utils/backend';
+
+export default {
+  props: {
+    failure: { type: Boolean },
+    hideIcon: { type: Boolean },
+    resolve: { type: Function, required: true },
+    error: { type: [String, Number], default: 'Something went wrong' },
```
davidyuk, line 36: In which case, Error can be a number? If somebody does like `throw 42`?

```suggestion
    error: { type: Error, required: true },
```

src/components/ErrorModal.vue
```diff
@@ -0,0 +1,140 @@
+<template>
+  <div
+    class="modal-mask"
+    @click="resolve"
+  >
+    <div class="modal-wrapper" @click.stop>
+      <div class="error-modal">
```
davidyuk, line 7: One of these wrappers seems to be extra

src/components/ErrorModal.vue
```diff
@@ -0,0 +1,140 @@
+<template>
+  <div
+    class="modal-mask"
```
davidyuk, line 3:
```suggestion
    class="error-modal"
```
proposing to name the root node the same as the component

src/components/ErrorModal.vue
```diff
@@ -0,0 +1,140 @@
...
+  position: fixed;
+  z-index: 9998;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  display: table;
+  background-color: rgba(0, 0, 0, 0.5);
+
+  .error-modal {
```
davidyuk, line 81: Let's use flex instead of `display: table`

```suggestion
  display: flex;
  background-color: rgba(0, 0, 0, 0.5);

  .error-modal {
    margin: auto;
```

src/components/ErrorModal.vue
```diff
@@ -0,0 +1,140 @@
...
+    text-align: center;
+    width: 25rem;
+    margin: 0 auto;
+    padding: 2.5rem 2.5rem 3.5rem 2.5rem;
+    background-color: $actions_ribbon_background_color;
+    border-radius: 0.25rem;
+    border: 1px solid $card_border_color;
+    box-shadow: 4px 4px 8px rgba(0, 0, 0, 0.25);
+    word-break: break-word;
```
davidyuk, line 90: why is this needed? User's input in the textarea, nothing should require to break words

src/components/ErrorModal.vue
```diff
@@ -0,0 +1,140 @@
...
+      margin-bottom: 0.85rem;
+    }
+
+    h1 {
+      color: $standard_font_color;
+      font-size: 1rem;
+      font-weight: 500;
+    }
+
+    .button {
```
davidyuk, line 108: There is the same button in AlertModal, don't you want to extract the common component?

src/utils/backend.js
```diff
@@ -2,7 +2,7 @@
 import store from '../store';
 import { wrapTry } from './index';
 
-const backendFetch = (path, ...args) => wrapTry(
+export const backendFetch = (path, ...args) => wrapTry(
```
davidyuk, line 5: Please define `errorreport` method here, instead of exporting `backendFetch`. Later we will replace all of them with API generated by swagger file.

src/components/ErrorModal.vue
```diff
@@ -0,0 +1,140 @@
...
+      float: right;
+    }
+  }
+
+  .report-text {
+    width: 100%;
+    background: #171717;
+    border-radius: 6px;
+    border: 1px solid #000;
+    resize: none;
```
davidyuk, line 127: I would do it vertically resizable, maybe with max-height.

src/App.vue
```diff
@@ -42,6 +43,15 @@ export default {
     ...mapState('aeternity', ['sdk']),
   },
   async created() {
+    window.addEventListener('unhandledrejection', (error) => {
```
davidyuk, line 46: Extract this handler to a new store plugin to don't waste `App.vue`. Also, add a handler for a usual exception as well.

src/components/ErrorModal.vue
```diff
@@ -0,0 +1,140 @@
...
+export default {
+  props: {
+    failure: { type: Boolean },
+    hideIcon: { type: Boolean },
+    resolve: { type: Function, required: true },
+    error: { type: [String, Number], default: 'Something went wrong' },
+  },
+  data() {
+    return {
+      SuccessIcon,
```
davidyuk, line 40: is it used somewhere?

src/utils/index.js
```diff
@@ -91,22 +91,20 @@ export const blockToDate = (goalBlock, height) => {
...
+      if (res.status !== 200) {
+        throw new Error(`Error code ${res.status}`);
+      }
+      if (!res.ok) throw new Error(4000);
       return res.json();
-    }).catch((error) => {
-      console.error(error);
-      return null;
     });
   } catch (err) {
```
davidyuk, line 107: You don't need an error handler if you are just passing an error through.

src/utils/index.js
```diff
@@ -91,22 +91,20 @@ export const blockToDate = (goalBlock, height) => {
...
-    if (store) { store.commit('setBackendStatus', false); }
-    return null;
+    throw new Error(err);
```
davidyuk, line 108: Also, we should expect that `err` is already an instance of Error/Error inheritor, why are you casting it to Error one more time?

src/utils/index.js
```diff
@@ -91,22 +91,20 @@ export const blockToDate = (goalBlock, height) => {
   return new Date(diff * 180000 + Date.now());
 };
 
-export const wrapTry = async (promise, store) => {
+// eslint-disable-next-line consistent-return
+export const wrapTry = async (promise) => {
```
davidyuk, line 95: remove `store` argument from its calls

src/utils/index.js
```diff
@@ -91,22 +91,20 @@ export const blockToDate = (goalBlock, height) => {
...
   try {
     return promise.then((res) => {
       if (!res) {
-        if (store) { store.commit('setBackendStatus', false); }
-        return null;
+        throw new Error('No response object');
```
davidyuk, line 99: In which case there can be no response object? I would remove this check as unnecessary

src/utils/index.js
```diff
@@ -91,22 +91,20 @@ export const blockToDate = (goalBlock, height) => {
...
       }
-      if (store) { store.commit('setBackendStatus', true); }
-      if (!res.ok) throw new Error(`Request failed with ${res.status}`);
+      if (res.status !== 200) {
+        throw new Error(`Error code ${res.status}`);
+      }
+      if (!res.ok) throw new Error(4000);
```
davidyuk, line 104: Why 4000?? I would just leave as it was before, `res.ok` check should be enough, and "Request failed with" looks better

davidyuk (commented):
please, write more accurate, ensure that it looks fine before asking for a review

src/App.vue
shapkarin: right, thanks

src/components/ErrorModal.vue
shapkarin: But how I will set https://github.com/aeternity/superhero-ui/blob/6a14ea86732f52fb0f947c3d62b0625128e39e9e/src/components/ErrorModal.vue#L49
the same lib used at the wallet

src/components/ErrorModal.vue
shapkarin: Will do

shapkarin:
@davidyuk thanks for the conversation I will move error handler to the Vue plugin and refact some. Will ping you soon.

src/store/modules/aeternity.js
```diff
@@ -114,11 +114,10 @@ export default {
           commit('setSdk', { instance });
           await dispatch('initTippingContractIfNeeded');
         }
-      } catch (err) {
-        commit('setBackendStatus', false, { root: true });
-        return;
+        return true;
+      } catch (error) {
+        return error;
```
davidyuk, line 119: who will handle the returned (not thrown) error? proposing to just remove `try`/`catch`

src/styles/_fonts.scss
```diff
@@ -4,3 +4,5 @@
 @import '~fontsource-ibm-plex-sans/500-normal.css';
 @import '~fontsource-ibm-plex-sans/700-normal.css';
 @import '~fontsource-ibm-plex-sans/700-italic.css';
+
+@import '~typeface-roboto/index.css';
```
davidyuk, line 8: @onvisions is it correct that we need to add a new font specially for this error popup?

src/utils/index.js
```diff
@@ -91,22 +91,20 @@ export const blockToDate = (goalBlock, height) => {
...
+      if (!res.ok) throw new Error(`Request failed with status ${res.status}`);
       return res.json();
-    }).catch((error) => {
-      console.error(error);
-      return null;
     });
   } catch (err) {
-    if (store) { store.commit('setBackendStatus', false); }
-    return null;
+    return err;
```
davidyuk, line 108: who will handle the returned error? Errors should be thrown!

src/utils/index.js
```diff
@@ -91,22 +91,20 @@ export const blockToDate = (goalBlock, height) => {
...
+// eslint-disable-next-line consistent-return
+export const wrapTry = async (promise) => {
   try {
     return promise.then((res) => {
-      if (!res) {
-        if (store) { store.commit('setBackendStatus', false); }
-        return null;
+      // if (!res) {
+      //   throw new Error('No response object');
+      // }
```
davidyuk, line 100:
```suggestion
```

src/utils/index.js
```diff
@@ -91,22 +91,20 @@ export const blockToDate = (goalBlock, height) => {
...
+      if (res.status !== 200) {
```
davidyuk, line 101: what is the difference between `res.status !== 200` and `!res.ok`?

src/utils/index.js
```diff
@@ -91,22 +91,20 @@ export const blockToDate = (goalBlock, height) => {
   return new Date(diff * 180000 + Date.now());
 };
 
-export const wrapTry = async (promise, store) => {
+// eslint-disable-next-line consistent-return
```
davidyuk, line 94: this method should return the server response or throw an error, I don't see reasons for inconsistency

src/components/ErrorModal.vue
```diff
@@ -0,0 +1,267 @@
...
+export default {
+  components: {
+    FailureIcon,
+    IconClose,
+    IconEye,
+    IconOK,
+  },
+  props: {
+    resolve: { type: Function, required: true },
+    error: { type: [Error, PromiseRejectionEvent], required: true },
```
davidyuk, line 88: why do you need the whole PromiseRejectionEvent, won't [`PromiseRejectionEvent.reason`](https://developer.mozilla.org/en-US/docs/Web/API/PromiseRejectionEvent/reason) be enough?

src/components/ErrorModal.vue
```diff
@@ -0,0 +1,267 @@
...
+        @click="resolve"
+      />
+      <template v-if="step === 1">
+        <FailureIcon />
+        <h1>{{ $t('reportBug.Title') }}</h1>
+        <h2>{{ $t('reportBug.SubTitle') }}</h2>
+        <p class="description">
+          {{ $t('reportBug.Description') }}
+        </p>
+        <div
```
davidyuk, line 18:
```suggestion
        <button
```
or
```suggestion
        <ButtonPlain
```

src/components/ErrorModal.vue
```diff
@@ -0,0 +1,267 @@
...
+  methods: {
+    async send() {
+      const report = {
+        appVersion: process.env.npm_package_version,
+        browser: detect(),
+        error: this.error,
+        platform: 'superhero-ui',
+        description: this.report,
+        // eslint-disable-next-line new-cap
+        time: new Date.now(),
```
davidyuk, line 106:
```suggestion
        time: Date.now(),
```

src/components/ErrorModal.vue
```diff
@@ -0,0 +1,267 @@
...
+    height: 40px;
+    margin-bottom: 15px;
+    padding: 0.5em 0.9em;
+    color: #fff;
+    font-size: 14px;
+    line-height: 24px;
+
+    &:focus {
+      background: #000;
+      border: 1px solid #1161fe;
```
davidyuk, line 220:
```suggestion
      border: 1px solid $secondary_color;
```

src/components/ErrorModal.vue
```diff
@@ -0,0 +1,267 @@
...
+      font-style: normal;
+      font-weight: 500;
+      font-size: 15px;
+      line-height: 24px;
+    }
+
+    .description {
+      font-size: 15px;
+      line-height: 24px;
+      color: #babac0;
```
davidyuk, line 177:
```suggestion
      color: $tip_note_color;
```

src/components/ErrorModal.vue
```diff
@@ -0,0 +1,267 @@
...
+      font-size: 0.75rem;
+      font-weight: 700;
+      justify-self: center;
+      line-height: 1.125;
+      padding: 0.65rem 1rem;
+      color: #fff;
+
+      &.cancel {
+        background: #232323;
+        color: #babac0;
```
davidyuk, line 199:
```suggestion
        color: $tip_note_color;
```

src/assets/APIError.svg
```diff
@@ -0,0 +1,9 @@
+<svg width="48" height="48" viewBox="0 0 48 48" fill="none" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
davidyuk, line 1:
```suggestion
<svg viewBox="0 0 48 48" fill="none" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
remove predefined sizes of other icons too

davidyuk (changes requested):
Not all issues of the previous review are resolved

src/store/modules/aeternity.js
shapkarin: right, I added that because of lint. will remove

src/utils/index.js
shapkarin: I guess that it needs `return throw err` because err is already an instance of Error. will add, thanks

src/components/ErrorModal.vue
shapkarin: maybe it's also just an Object, because I also pass there an whole error and also stack?

src/utils/index.js
shapkarin: There is no difference 
```
The ok read-only property of the Response interface contains a Boolean stating whether the response was successful (status in the range 200-299) or not.
```

src/utils/index.js
shapkarin: I got this lint issue https://eslint.org/docs/rules/no-useless-catch

src/components/ErrorModal.vue
shapkarin: changed, but it's for other line

shapkarin: https://github.com/aeternity/superhero-ui/blob/5904e141e1ffb4d6b751d74f8df6545acd3174ac/src/components/ErrorModal.vue#L18

src/store/plugins/errorHandler.js
```diff
@@ -1,19 +1,9 @@
 export default ({ dispatch }) => {
-  const errors = [];
   const handlerBody = async (error) => {
-    try {
-      if (errors.lenght === 0) {
-        console.log({ errors });
-        dispatch('modals/open', {
```
shapkarin: await...
don't need try catch because there is not rejection

src/components/Button.vue
```diff
@@ -0,0 +1,35 @@
+<template>
+  <button
+    :disabled="disabled"
+    class="button"
+    @click="$emit('click', $event)"
```
davidyuk, line 5:
```suggestion
    class="button"
    v-bind="$attrs"
    v-on="$listeners"
```

shapkarin: @davidyuk why it can't be disabled?

src/store/plugins/errorHandler.js
```diff
@@ -0,0 +1,25 @@
...
+    if (error) errors.unshift(error);
+    if (invoke && errors.length !== 0) {
+      invoke = false;
+      await dispatch('modals/open', {
+        name: 'error',
+        error: errors[0],
+      });
+      errors.shift();
+      invoke = true;
+      handlerBody();
```
davidyuk, line 14: prefer cycle instead of recursion because it is easier to read

src/store/plugins/errorHandler.js
```diff
@@ -0,0 +1,25 @@
...
+  let invoke = true;
+  const handlerBody = async (error) => {
+    if (error) errors.unshift(error);
+    if (invoke && errors.length !== 0) {
+      invoke = false;
+      await dispatch('modals/open', {
+        name: 'error',
+        error: errors[0],
+      });
+      errors.shift();
```
davidyuk, line 12:
```suggestion
        error: errors.shift(),
      });
```

src/store/plugins/errorHandler.js
```diff
@@ -0,0 +1,25 @@
+export default ({ dispatch }) => {
+  const errors = [];
+  let invoke = true;
+  const handlerBody = async (error) => {
+    if (error) errors.unshift(error);
+    if (invoke && errors.length !== 0) {
```
davidyuk, line 6:
```suggestion
    if (!invoke || errors.length === 0) return;
```
proposing to avoid extra nesting

src/components/ErrorModal.vue
```diff
@@ -0,0 +1,267 @@
...
+    .button {
+      background-color: $secondary_color;
+      border: none;
+      border-radius: 0.25rem;
+      font-size: 0.75rem;
+      font-weight: 700;
+      justify-self: center;
+      line-height: 1.125;
+      padding: 0.65rem 1rem;
+      color: #fff;
```
davidyuk, line 191: Isn't these tyles already defined in Button component?

src/components/ErrorModal.vue
```diff
@@ -0,0 +1,267 @@
...
+    FailureIcon,
+    IconClose,
+    IconEye,
+    IconOK,
+    ButtonPlain,
+    Button,
+  },
+  props: {
+    resolve: { type: Function, required: true },
+    error: { type: [Object, Error, PromiseRejectionEvent.reason], required: true },
```
davidyuk, line 90: `PromiseRejectionEvent.reason` is not a type

package.json
```diff
@@ -36,7 +37,7 @@
     "lodash-es": "^4.17.20",
     "sophia-bonding-curve": "github:aeternity/BondingCurve#1.0.0-alpha",
     "soundcloud-widget": "^0.2.1",
-    "tipping-contract": "github:aeternity/tipping-contract#v3.0.1-alpha1",
+    "tipping-contract": "github:aeternity/tipping-contract#v3.0.0-alpha1",
```
davidyuk, line 40: why have you downgraded this package?

src/utils/backend.js
```diff
@@ -169,3 +173,13 @@ export default class Backend {
     headers: { 'Content-Type': 'application/json' },
   });
 }
+
+export const issueReport = (report) => {
```
davidyuk, line 177: won't be consistent to define it as a static method of the above class?

src/utils/index.js
davidyuk: this one doesn't looks like resolved

src/utils/index.js
```diff
@@ -91,22 +91,15 @@ export const blockToDate = (goalBlock, height) => {
...
+  // eslint-disable-next-line no-useless-catch
   try {
     return promise.then((res) => {
-      if (!res) {
-        if (store) { store.commit('setBackendStatus', false); }
-        return null;
-      }
-      if (store) { store.commit('setBackendStatus', true); }
-      if (!res.ok) throw new Error(`Request failed with ${res.status}`);
+      if (!res) throw new Error('Responce object is undefined');
```
davidyuk, line 98: remove `.catch((err) => console.error(err))` in https://github.com/aeternity/superhero-ui/blob/8650d831804fe2e98e7862aa3de0109605bc78a2/src/utils/backend.js#L6 instead

src/assets/iconCloseRebranded.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="14" height="14" viewBox="0 0 14 14" fill="none" xmlns="http://www.w3.org/2000/svg">
```
davidyuk, line 1: can you remove icon size here as well?

src/assets/iconEye.svg
```diff
@@ -0,0 +1,6 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
davidyuk, line 1: can you remove icon size here as well?

src/components/ErrorModal.vue
```diff
@@ -0,0 +1,267 @@
...
+  },
+  data() {
+    return {
+      report: '',
+      step: 1,
```
davidyuk, line 95: I would name it `isSent` or `isSuccess` having a boolean value instead, we don't have any multiple steps

src/components/ErrorModal.vue
```diff
@@ -0,0 +1,267 @@
...
+
+    .description {
+      font-size: 15px;
+      line-height: 24px;
+      color: $tip_note_color;
+    }
+
+    .reportTitle {
+      font-size: 15px;
+      color: #787878;
```
davidyuk, line 178:
```suggestion
      color: $light_font_color;
```

src/components/ErrorModal.vue
```diff
@@ -0,0 +1,267 @@
...
+          class="detailsButton"
+          @click="toggleDetails"
+        >
+          <IconEye />
+          {{ $t('reportBug.ViewDetails') }}
+        </ButtonPlain>
+        <p v-show="details">
+          {{ JSON.stringify(error) }}
+        </p>
+        <h2 class="reportTitle">
```
davidyuk, line 28:
```suggestion
        <h2 class="report-title">
```
usually, we are using kebab-case for class names, why you have switched to camelCase here and in other places?

src/utils/backend.js
```diff
@@ -116,7 +116,11 @@ export default class Backend {
     if (ordering) queryParams.set('ordering', ordering);
     if (direction) queryParams.set('direction', direction);
     if (search) queryParams.set('search', search);
-    return backendFetch(`tokenCache/wordRegistry?${queryParams.toString()}`);
+    return backendFetch(`tokenCache/wordRegistry?${queryParams.toString()}`)
+      .catch((err) => {
+        console.log(`tokenCache/wordRegistry response with: ${err}`);
+        return [];
+      });
```
davidyuk, line 123: proposing to move it to where it is called because this code shouldn't be responsible for error handling

davidyuk (commented):
why you still didn't rebase it?

src/utils/index.js
shapkarin: resolved thanks

src/store/plugins/errorHandler.js
```diff
@@ -0,0 +1,40 @@
...
+      if (invoke && errors.length !== 0) {
+        invoke = false;
+        await dispatch('modals/open', {
+          name: 'error',
+          error: errors.shift(),
+        });
+        invoke = true;
+      }
+    }
+  };
```
shapkarin, line 30: @davidyuk can you have a look at handlerBodyLoop function?

davidyuk: I would do it this way:
```suggestion
  const handlerBodyLoop = async (error) => {
    errors.unshift(error);
    if (!invoke) return;
    invoke = false;
    while (errors.length) {
      // eslint-disable-next-line no-await-in-loop
      await dispatch('modals/open', {
        name: 'error',
        error: errors.shift(),
      });
    }
    invoke = true;
  };
```

src/utils/backend.js
shapkarin: Why we can't later to wrapp the api methods with the same wrapTry function to catch all errors at one place? Like 
```js
backendFetch = (method) => wrapTry(sdk.api[method])
```

package.json
shapkarin: I got that after merge with develop, will back v3.0.1-alpha1

src/store/plugins/errorHandler.js
shapkarin: will keep that in mind

src/components/ErrorModal.vue
shapkarin: done

src/utils/backend.js
shapkarin: right, done

src/utils/index.js
shapkarin: @davidyuk so now the error is thrown automatically by the fetch promise and handled next at the plugin

src/App.vue
shapkarin: great thanks a lot for that catch! use plugin is much better

src/utils/index.js
shapkarin: thanks, done

src/assets/iconCloseRebranded.svg
shapkarin: done

src/utils/backend.js
shapkarin: @davidyuk currently it used by 4 files https://github.com/aeternity/superhero-ui/search?q=Backend.getWordRegistry and no one of that fetch is works fine, that's the reason why I defined it here. And the return empty array is currently for that like https://github.com/aeternity/superhero-ui/blob/ddaeb71bc9c2fb9886d9eafeef989b57a1f28a13/src/store/index.js#L57

davidyuk: > Why we can't later to wrapp the api methods with the same wrapTry function to catch all errors at one place? 

because it is a bad practice

3 or 4 use cases are not relevant, fix only the call in App.vue

src/utils/index.js
davidyuk: You don't need an error handler if you are just passing an error through.

src/utils/index.js
davidyuk:
```suggestion
```

shapkarin:
@davidyuk maybe I also should to remove all other catch? https://github.com/aeternity/superhero-ui/search?q=catch%28

shapkarin:
@mradkov can you please have a look at the PR while @davidyuk is not available. The main goal of that PR is to show a popup with the trace stack for any errors in the app and if there is more than one error user will see the next one popup after hi will finish with the previous one. thanks.

src/components/ErrorModal.vue
```diff
@@ -0,0 +1,264 @@
...
+  },
+  methods: {
+    async send() {
+      const report = {
+        appVersion: process.env.npm_package_version,
+        browser: detect(),
+        error: this.error,
+        platform: 'superhero-ui',
+        description: this.report,
+        time: Date.now(),
```
davidyuk, line 107: maybe more useful to remember the time when the modal is opened to don't record how long the user need to fill the description and to make easily to find the corresponding request to the backend (in case if error popup arrear because of backend response)

src/components/UserInfo.vue
```diff
@@ -395,8 +394,7 @@ export default {
           this.profile.biography = this.profile.biography || '';
           this.profile.coverImage = this.profile.coverImage || '';
           this.$store.commit('setUserProfile', profile);
-        })
-        .catch(console.error);
+        });
```
davidyuk, line 397: won't it throw an error for a newly created aeternity account?

src/utils/backend.js
```diff
@@ -1,13 +1,10 @@
-// eslint-disable-next-line import/no-cycle
```
davidyuk, line 1: this should allow removing the same in related files as well

src/utils/backend.js
```diff
@@ -1,13 +1,10 @@
...
 import { wrapTry } from './index';
 
 const backendFetch = (path, ...args) => wrapTry(
-  fetch(`${process.env.VUE_APP_BACKEND_URL}/${path}`, ...args).catch((err) => console.error(err), store),
+  fetch(`${process.env.VUE_APP_BACKEND_URL}/${path}`, ...args),
 );
 
-const backendFetchNoTimeout = (path, ...args) => fetch(`${process.env.VUE_APP_BACKEND_URL}/${path}`, ...args)
-  .catch((err) => console.error(err));
+const backendFetchNoTimeout = (path, ...args) => fetch(`${process.env.VUE_APP_BACKEND_URL}/${path}`, ...args);
```
davidyuk, line 7: current version of `wrapTry` doesn't have any timeouts, you can use `backendFetch` instead of `backendFetchNoTimeout`

davidyuk (commented):
in general looks much better, except for issues from below there are still two unresolved issues from the previous reviews
@shapkarin I have asked you to solve them several times, but seems that you still didn't get me, write me if you have any questions, don't mark them as resolved

src/components/ErrorModal.vue
shapkarin: @davidyuk I will move `Date.now()` outside of the Vue component, and add it to the plugin.

src/utils/index.js
shapkarin: @davidyuk I removed the whole try catch and let the error throw without that definition

src/App.vue
```diff
@@ -90,7 +94,9 @@ export default {
         this.$store.dispatch('updatePinnedItems'),
         this.$store.dispatch('updateUserProfile'),
         (async () => {
-          const balance = await this.sdk.balance(this.address).catch(() => 0);
+          const balance = await this.sdk.balance(this.address)
```
thepiwo, line 97: `getBalance defaults to 0

src/assets/iconOK.svg
```diff
@@ -0,0 +1,9 @@
+<svg viewBox="0 0 48 48" fill="none" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+<rect width="48" height="48" fill="url(#pattern0)"/>
+<defs>
+<pattern id="pattern0" patternContentUnits="objectBoundingBox" width="1" height="1">
+<use xlink:href="#image0" transform="scale(0.0208333)"/>
+</pattern>
+<image id="image0" width="48" height="48" xlink:href="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAADAAAAAwCAYAAABXAvmHAAABe0lEQVRoge3YsWqDQBgH8EsiGCtaxc3JkjWQyS1pp9BJMkogUFdDX8CMmYSsASdXX8Q3yBvkDfIGX4eCJG2+U4Lt3Qc3/DeV/+/gOz2Z53lAOUx0AQUQXUABRBdQANEFFOCvHuwGPhh5DIPLERiUwKAErc7ADXwaACOPm+LX0atUfoAb+Dcr/zPPr1O5AeMiQcszKMGOQnkBzmzCLS89QKuzVoAzm8gJsDaL1vIMyl4XDAU40xdu7t0zPB/kADztVu35fL+5B9s25QXsVs31bdvmdUanvXwAOwo7r/7otAc7CpuY2yUYedxEq7Nf0asUfYP/O+DRGHlMGzC4HGkDhucDbcC4SGgDzO2SNgD7giUDkOY98Ei0OqMNwAaYDAAbYDIA3hGUBAArTwLAG2ASAN4AkwDwBpgEoO0fkvQAXvneAJ73/T9oXCSdD/Z9DDAKsNZzMD/euLHWc/Sh1mYBRh6DXqV3j4hdgp3AOgEoRQFERwFERwFERwFE5wumYmZgGwDXKAAAAABJRU5ErkJggg=="/>
```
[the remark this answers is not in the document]
thepiwo: :D

src/App.vue
shapkarin: thanks, I remember that revert it on place, forgot about the second one

shapkarin: @thepiwo it's already done, please check the next line with comment

Liubov-crypto:
When I have clicked on the bug description field - the popup disappeared and I was not able to provide additional information:
retip on the feed

After clicking on the Submit button, there is no notification that my report has been sent. 
Also, if I click "View full technical details to be submitted" nothing happens.

shapkarin:
Thanks. I guess that becomes after the refactoring, will fix that tomorrow.

Liubov-crypto:
> Thanks. I guess that becomes after the refactoring, will fix that tomorrow.

Welcome! 
I couldn't get any other errors besides this one, during performing transactions with tip, retip, comments, vote, revoke votes, etc.

shapkarin:
@Liubov-crypto fixed with last commit. @davidyuk there is issue with the `@click.exact` approach, looks like there is exact needs to be added for each event so I just revert to the `@click.stop`

davidyuk (commented):
I have tried to throw an error from a click event handler, but it wasn't caught by `window.onerror` (https://github.com/vuejs/vue/issues/5901), we need to catch these errors as well.

src/App.vue
thepiwo: using `getBalance` defaults to 0, my comment doesn't seem to be resolved

src/assets/iconOK.svg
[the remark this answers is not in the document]
thepiwo: @shapkarin @onvisions what to do with that?

thepiwo (commented):
why is rejecting my consent a backend error that has to be reported?

image

thepiwo (commented):
how is submitting that error helpful?

image

thepiwo (commented):
why is there no error when rejecting to sign a comment?

davidyuk:
> how is submitting that error helpful?

do you mean that the requested URL is missed? The stacktrace should be included in the report and using it we should be able to determine on which line exception was raised. Also, ":1:55572" should be recoverable using the corresponding version of the source code.

shapkarin:
@thepiwo @davidyuk I can recheck that all after time off. Or you can proceed without me.

src/components/ErrorModal.vue
```diff
@@ -0,0 +1,264 @@
+<template>
+  <div
+    class="error-modal"
+    @click="resolve"
+  >
+    <div class="modal-body" @click.stop>
```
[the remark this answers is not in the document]
shapkarin: right, it's .self not .exact

src/assets/iconOK.svg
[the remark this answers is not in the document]
onvisions: I expected @shapkarin to ask me for animated SVG. Figma doesn't support animated SVGs so in FIgma file the animated icons are GIFs but I have said that I would create the animated vector versions.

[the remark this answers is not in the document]
onvisions: I'm not creating a detailed design review but the implementation completely lacks the micro interactions from the Figma prototype. I mean hover effects over buttons, icons, close X button, input fields. Hover and focused states have different colors and styles applied according to design. I don't think it's much more work to implement. 
Also from UX point of view the field for bug description should expand several more lines. Its max-length should be at least 120px to look more like a regular text area. 
The "eye" icon for expand/collapse is **replaced** in latest design by **chevron up/down** after feedback and product request. It's updated in Figma too.
Please check again the size of icons to be sure they match dimensions set in the prototype.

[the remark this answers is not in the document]
onvisions: SVG animated resources to replace the GIF icons are ready. I'm not sure if I should add or link it here so please sync in PM (whoever will be working to finalize this feature) so I can sent it to you.

davidyuk:
The current version only introduces a global error handler and a modal to send reports.

You can throw some errors here: https://test-errors.aggregator.z52da5wt.xyz/

Liubov-crypto (commented):
Tested, in general LGTM. 
During testing, I received an Unhandled Error pop-up when canceling the signing of a transaction, but I was unable to throw other types of pop-ups describing the error.

Also, I can see this at the top of the feed. Is this correct? 
errors

davidyuk:
>  I received an Unhandled Error pop-up when canceling the signing of a transaction

This is not the right behaviour, I have applied one commit from #1133 and some additional error handling.

@Liubov-crypto Try to find additional cases when the error report modal appears. We need to fix all of them.

> Also, I can see this at the top of the feed. Is this correct?

This is for testing purposes, these buttons added in a separate branch based on this one.

Liubov-crypto:
> > I received an Unhandled Error pop-up when canceling the signing of a transaction
> 
> This is not the right behaviour, I have applied one commit from #1133 and some additional error handling.
> 
> @Liubov-crypto Try to find additional cases when the error report modal appears. We need to fix all of them.
> 
> > Also, I can see this at the top of the feed. Is this correct?
> 
> This is for testing purposes, these buttons added in a separate branch based on this one.
> 
> link to designs: https://www.figma.com/file/OrPoPfiEYfXh7pOaVVYOaJ/SH-Website-Rebranded?node-id=3144%3A0

During testing, I noticed that the error notification pop-up only appears when I decline to sign a transaction. All the pop-ups I got look like this. Only the technical description is different:
укк

Also, the following test cases where I rejected the transaction but did not receive the modal with an error, but saw errors on the feed:
1.  Cancel posting a comment
2. Cancel send tip on the feed transaction
3. Cancel send post without tipping on the feed
4. Cancel Tip comment transaction 
5. Cancel off/on Youtube/Soundcloud cookies 

This is the information from the console when I declined signing the transaction:
cons
sign transaction rejection

I tried to disconnect the internet connection and then restart the page, but the error message pop-up did not appear.

AtanasKrondev (approved):

PR #1029, opened 2021-03-04
fix: adjusts some backend requests to comply with stricter standards

kenodressel:
In preparation for the https://github.com/aeternity/tipping-community-backend/pull/277 which will be deployed in 1.9.0

### DIFF
--- a/src/utils/backend.js
+++ b/src/utils/backend.js
@@ -111,7 +111,13 @@ export default class Backend {
 
   static getTokenInfo = async () => backendFetch('tokenCache/tokenInfo');
 
-  static getWordRegistry = async (ordering, direction, search) => backendFetch(`tokenCache/wordRegistry?${ordering ? `ordering=${ordering}` : ''}${direction ? `&direction=${direction}` : ''}${search ? `&search=${search}` : ''}`);
+  static getWordRegistry = async (ordering, direction, search) => {
+    const queryParams = new URLSearchParams();
+    if(ordering) queryParams.set('ordering', ordering);
+    if(direction) queryParams.set('direction', direction);
+    if(search) queryParams.set('search', search);
+    return backendFetch(`tokenCache/wordRegistry?${queryParams.toString()}`);
+  }
 
   static getWordSaleVotesDetails = async (address) => backendFetch(`tokenCache/wordSaleVotesDetails/${address}`);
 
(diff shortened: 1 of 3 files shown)


src/utils/backend.js
```diff
@@ -111,7 +111,7 @@ export default class Backend {
 
   static getTokenInfo = async () => backendFetch('tokenCache/tokenInfo');
 
-  static getWordRegistry = async (ordering, direction, search) => backendFetch(`tokenCache/wordRegistry?ordering=${ordering}&direction=${direction}&search=${search || ''}`);
+  static getWordRegistry = async (ordering, direction, search) => backendFetch(`tokenCache/wordRegistry?${ordering ? `ordering=${ordering}` : ''}${direction ? `&direction=${direction}` : ''}${search ? `&search=${search}` : ''}`);
```
davidyuk, line 114: Without `ordering` it will construct the query like `wordRegistry?&direction...`, it still may work well, but I would propose to construct it using URL class. Or finally, generate this API by swagger file then we won't be needed to deal with it.

kenodressel: We could start working on the Swagger API generation. The swagger spec should be fairly accurate now (at least regarding the requests). But for now we only need these two minor changes to avoid params sending the value `undefined`. I tried to keep the changes in the one-line fashion but I am happy to adjust them using the URL class.

src/utils/backend.js
```diff
@@ -111,7 +111,13 @@ export default class Backend {
 
   static getTokenInfo = async () => backendFetch('tokenCache/tokenInfo');
 
-  static getWordRegistry = async (ordering, direction, search) => backendFetch(`tokenCache/wordRegistry?ordering=${ordering}&direction=${direction}&search=${search || ''}`);
+  static getWordRegistry = async (ordering, direction, search) => {
+    const queryParams = new URLSearchParams();
+    if(ordering) queryParams.set('ordering', ordering);
+    if(direction) queryParams.set('direction', direction);
+    if(search) queryParams.set('search', search);
```
davidyuk, line 118:
```suggestion
    if (ordering) queryParams.set('ordering', ordering);
    if (direction) queryParams.set('direction', direction);
    if (search) queryParams.set('search', search);
```
this should fix linter errors: https://github.com/aeternity/superhero-ui/pull/1029/checks?check_run_id=2038213738#step:7:1098

davidyuk (approved):

mradkov (approved):

PR #1030, opened 2021-03-05
Price history chart

AtanasKrondev:


thepiwo:
- design seems broken (lct3)
image

 - chart should probably be rendered after data is fetched
 - tab bar design looks weird

mradkov:
Screenshot 2021-03-08 at 9 13 27

I still see some small miss-alignments. Can you fix those?

onvisions:
A few comments by me:

Screenshot_4

Screenshot_5

I noticed also that some styles that have been applied are now missing again as if it's an older version. 
See the Sell and Buy hover, the tabs text color and background. The search box (search for tokens) ... @AtanasKrondev we discussed some changes and I remember you had applied them but now I can't see the fixes from this link above.

thepiwo (approved):
looks good to me

mradkov (approved):

PR #1041, opened 2021-03-24
WordBazaar Deeplink interactions

AtanasKrondev:
Pass transactions through deeplinks to the wallet

Wallet PR: https://github.com/aeternity/superhero-wallet/pull/930


src/store/modules/aeternity.js
```diff
@@ -261,32 +261,46 @@ export default {
...
     },
     async tokenSaleMethod(
-      { dispatch },
+      { dispatch, state: { useSdkWallet } },
       {
         contractAddress,
         method,
         args = [],
         options = {},
+        withDeepLink = false,
```
davidyuk, line 270: is this param for testing purposes? I don't see reasons to use deeplinks if a sdk-based wallet is available

davidyuk (commented):
so far looks fine, interesting to see how you will handle the cases when WB needed to show something if transaction complete/rejected

AtanasKrondev:
> so far looks fine, interesting to see how you will handle the cases when WB needed to show something if transaction complete/rejected

And here I'm stuck. I managed to sign a send transaction by the transaction hash. But I can not manage to get the decodedResult that normally is returned by executing a contract method.
Like [here](https://github.com/aeternity/superhero-ui/blob/develop/src/store/modules/aeternity.js#L309) it's needed in order to proceed further.

@davidyuk can you suggest how it can be done?

davidyuk:
I have answered this in PM before:

> if broadcasting on UI side you should send it to blockchain, wait until it is mined and get tx hash
otherwise wallet should return tx hash
using tx hash try to recover decoded result using getTxInfo:
https://github.com/aeternity/aepp-sdk-js/blob/584eb5e4ff875a9d8dca71e4492076d58f288cc3/es/ae/contract.js#L49

thepiwo:
@AtanasKrondev @kenodressel @davidyuk whats the state of this?

AtanasKrondev:
> @AtanasKrondev @kenodressel @davidyuk whats the state of this?

What is the state of the need of implementation deeplinks in general and should we proceed?

kenodressel:
Good question, can @davidyuk answer this? Else we should probably do a session around this with the wallet team.

PR #1043, opened 2021-03-29
FAQ two questions added.

aniputapet:
Waiting on instructions how to add links, and then will proceed with all FAQ changes. Moreover, need to see how the current changes look - maybe is to much text

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -87,12 +87,12 @@
       "subheader": "ABOUT SUPERHERO.com",
       "questions": [
         {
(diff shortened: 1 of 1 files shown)


thepiwo:
@aniputapet @mradkov is this still needed?

davidyuk:
As I got, links are not needed in the nearest release, and @aniputapet can test changes locally. We are waiting for the latest wording updates from her.

davidyuk:
replaced by #1050

PR #1040, opened 2021-03-24
Rebrand tip icon

shapkarin:

PR #1045, opened 2021-04-01
Refactor TipInput, remove unused code

davidyuk:
made after checking #1040


src/components/TipInput.vue
```diff
@@ -4,15 +4,11 @@
...
       class="button"
+      :class="{ tipped: tipUrlStats.isTipped }"
       :title="title"
       @click="useSdkWallet && (showModal = true)"
     >
-      <template v-if="userAddress">
-        <IconTipUser />
-      </template>
-      <template v-else>
-        <IconTip :class="{ tipped: tipUrlStats.isTipped }" />
```
shapkarin, line 14: it's has states at the UI-kit
Снимок экрана 2021-04-01 в 20 54 45

src/components/TipInput.vue
```diff
@@ -99,7 +94,6 @@ export default {
     Modal,
     Dropdown,
     IconTip,
-    IconTipUser,
```
shapkarin, line 102: agree is @onvisions will approve that

davidyuk: actually, I don't have access to the latest designs of user profile page (can somebody share a link?), but as I see at #1040 they were already broken

src/components/TipInput.vue
davidyuk: yep, I think is better to apply hover designs when hovered on the whole button, not only the part of it (the tip icon)

shapkarin: It has 4 states. True, it can be done with change class at the whole button, but at this PR there is no any styles for the icon background color.

davidyuk: have you checked the deployed version? https://feature-refactor-1040.aggregator.z52da5wt.xyz/

shapkarin: @davidyuk checked but without auth the wallet. looks like it has all states.

shapkarin (approved):

PR #1049, opened 2021-04-08
Minor design adjustments

AtanasKrondev:
- "My Profile" position in navigation changed to 2nd
- When on "Feed" -> "Main" the option is "Create post" instead of "Send Tip"
- Not working buttons (functionality not yet implemented) hidden in "Create Post" and "Tip Record"
- Iframe wallet height adjusted to fit with the re-design

### DIFF
--- a/src/components/layout/RightSectionWallet.vue
+++ b/src/components/layout/RightSectionWallet.vue
@@ -175,7 +175,7 @@ export default {
     display: block;
     border: none;
     width: 100%;
-    height: 470px;
+    height: 600px;
     transition: height 0.25s;
 
     &.half-closed {
--- a/src/components/layout/RightSectionWallet.vue
+++ b/src/components/layout/RightSectionWallet.vue
@@ -161,13 +161,13 @@ export default {
 
 <style lang="scss" scoped>
 .right-section-wallet {
-  padding: 0.8rem 1rem;
+  padding: 8px 16px;
 
   &.iframe {
     padding: 0;
 
     .right-section-title {
-      padding: 0.8rem 1rem;
+      padding: 8px 16px;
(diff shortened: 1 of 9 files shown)


src/components/layout/sendTip/SendPost.vue
```diff
@@ -32,15 +32,17 @@
...
-        <ButtonPlain disabled>
-          <IconEmoji />
-        </ButtonPlain>
-        <ButtonPlain disabled>
-          <IconPoll />
-        </ButtonPlain>
-        <ButtonPlain disabled>
-          <IconThreeDots />
-        </ButtonPlain>
+        <template v-if="UNFINISHED_FEATURES">
```
davidyuk, line 35: Proposing to mark this as unfinished instead of commenting to keep code highlighting, simplify refactoring of related parts and so on.
We have been using this approach before in Base aepp: https://github.com/aeternity/aepp-base/blob/develop/.env.production#L1

src/components/layout/RightSectionWallet.vue
```diff
@@ -175,7 +175,7 @@ export default {
     display: block;
     border: none;
     width: 100%;
-    height: 470px;
+    height: 600px;
```
davidyuk, line 178: I didn't get why we have been asked to do it. The only problem that I see is that "Terms Privacy  Branded by Venture" is hidden and can't be scrolled to at 1920x1080 screen:
Screenshot 2021-04-09 at 02 08 10

etharner: It's because the wallet dimensions have changed, now it has 600px height.

davidyuk (approved):

PR #1067, opened 2021-04-27
Remove URLS and numbers from wording

### LINKED ISSUE #1035: Don't store URLs in localisation files
https://github.com/aeternity/superhero-ui/blob/7dd16a2f218d3e55bef88f1a8e30f6d3fa9790ae/src/locales/es.json#L654-L659

They should not be "translated" to multiple languages, this is an error-prone approach.

The correct way would be something like
```json
    "GetAe": {
      "Platforms": {
        "jelly": "BTC/{ae} ETH/{ae}",
        "coinmarketcap": "Discover more {ae} exchanges",
        "faucet": "Testnet {ae} tokens",
      }
    },
```
cc: @AtanasKrondev

### LINKED ISSUE #1037: Don't use numbers as translation keys for page titles
https://github.com/aeternity/superhero-ui/blob/52c3d48d0535319b2025e4d1b4d50dd15f242fc0/src/locales/en.json#L485-L490

- adding a new page makes to update indexes of all other pages
- if we will have a merge conflict it won't be impossible to solve it without Google Translate

I think it is fine to use arrays in localization if it represents something not explicitly referenced in the code. Like a set of paragraphs or steps, but the code shouldn't contain a specific icon for each step, otherwise, we should use words as keys.

AtanasKrondev:
Fixes #1035 
Fixes #1037

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -660,20 +660,11 @@
       ]
     },
     "GetAe": {
-      "Platforms": [
-        {
-          "Text": "BTC/{ae} ETH/{ae}",
-          "Link": "https://jelly.market/"
-        },
-        {
-          "Text": "Discover more {ae} exchanges",
-          "Link": "https://coinmarketcap.com/en/currencies/aeternity/"
-        },
-        {
-          "Text": "Testnet {ae} tokens",
-          "Link": "https://faucet.aepps.com/"
-        }
-      ]
+      "Platforms": {
+        "Jelly": "BTC/{ae} ETH/{ae}",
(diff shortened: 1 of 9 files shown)


thepiwo:
@AtanasKrondev @kenodressel whats the state of this?

AtanasKrondev:
> @AtanasKrondev @kenodressel whats the state of this?

I propose merging

kenodressel:
why are we migrating away from a configurable solution? the previous state looked better to me

kenodressel (approved):

thepiwo (approved):

PR #1069, opened 2021-04-28
Enable account switching

CedrikNikita:


thepiwo:
@kenodressel @CedrikNikita whats the state of this?

Issue #1028, opened 2021-03-04
Avoid wording depending on the screen size


davidyuk:
@onvisions in the latest WordBazaar I saw the [wording](https://github.com/aeternity/superhero-ui/pull/1023#discussion_r587286574) that depends on the screen size. I'm proposing to avoid that, because:
- it is tricky to implement using straightforward CSS
- this complicates internationalization

In general a tunning of wording into available space is a fragile approach. I think all wording should be placed with available free space around to simplify internationalization. For example, we were translating the Base aepp interface into Russian and most wording was wider than in English. Would be best if designs consider it at the beginning.

onvisions:
I've got your point about resolution-specific wording. I'm thinking on another solution for navigation on small screens. It will be a quick fix to decrease font-size and make the whole tab bar horizontally swipeable under certain resolution (when it's overflowing the screen size).  However for pixel perfect solution I will need some more time to think and create mockups.

PR #1073, opened 2021-04-29
chore: switch to a wallet with new sdk

davidyuk:


Liubov-crypto:
When creating a new wallet using iframe:
word

Different names in UI and wallet. Names are not displayed in the wallet. They disappear after registration:
different names in ui and wallet

Error in iframe:
ok err in ff

Scan error (although I have allowed the use of the camera):
scan error
camera

Create an invite link using FT:
inv link with FT

Names in notifications shown as: 
names in notifications

if you overload UI several times, then the wallet will be logged out.

These errors also remain #1028 #907 (comment)

davidyuk:
> When creating a new wallet using iframe

cc: @etharner 

> Scan error

The architectural problem, need a special guide

> Create an invite link using FT

As I know, we decided to disallow it

> Names in notifications shown as

Should be fixed in develop

Should be fixed on wallet side:
- Different names in UI and wallet. Names are not displayed in the wallet.
- OK error

Liubov-crypto:
> > When creating a new wallet using iframe
> 
> cc: @etharner
> 
> > Scan error
> 
> The architectural problem, need a special guide for the end-user
> 
> > Create an invite link using FT
> 
> As I know, we decided to disallow it
> 
> > Names in notifications shown as
> 
> Should be fixed in develop
> 
> Should be fixed on wallet side:
> 
> * Different names in UI and wallet. Names are not displayed in the wallet.
> * OK error



As I know, we decided to disallow it - yes, Ivelin fixed his in pr. 

Names in notifications shown as - fixed

Should be fixed on the wallet side:

Different names in UI and wallet. Names are not displayed in the wallet. - fixed
OK error - fixed

Scan error -scan QR code work in https://feature-update-sdk.wallet.z52da5wt.xyz/payments/

but then a black screen appears and the bottom menu becomes top:
errrr

Issue #740, opened 2020-07-28
What is the status of Bootstrap in this project?


davidyuk:
How actively it is planned to use it?

Removing of bootstrap-vue in #682 looks like that Bootstrap is more in the deprecated state, is it true?

From my perspective, it makes styles messier especially when Bootstrap styles overrides: https://github.com/aeternity/superhero-ui/blob/bda0e1d51f004d8a527063c35d7f7d5595f19938/src/components/SendComment.vue#L109-L110

Also, I can't use short and semantic classes like `modal` or `row` because they overlap with global bootstrap styles.

mradkov:
> Removing of bootstrap-vue in #682 looks like that Bootstrap is more in the deprecated state, is it true?

It's in a deprecated state. Though there is a bootstrap dependency in the Time-back machine date picker (maybe we can refactor it to use custom or some other pickers).

Issue #1112, opened 2021-06-28
Go back button affects iframe wallet


Liubov-crypto:
https://user-images.githubusercontent.com/69896204/123628548-e3a6ad00-d81b-11eb-8688-ea9668876f24.mov

The same behavior on develop branch.

Issue #1111, opened 2021-06-28
Different .chain names in wallet and user interface


Liubov-crypto:
I found that the .chain name in the UI and in the wallet is not synced. The user interface displays the name from the sub-account, in the wallet name from the main account:

![different chain names in wallet and UI](https://user-images.githubusercontent.com/69896204/123628028-58c5b280-d81b-11eb-9076-cda3575ccf12.png)

The same on develop branch.

PR #1114, opened 2021-06-30
Refactor styles

davidyuk:
fixes partially #740


src/components/tipRecords/TipMedia.vue
```diff
@@ -1,106 +1,91 @@
...
+      });
+    },
+  },
 };
 </script>
 
 <style lang="scss" scoped>
 .tip-media {
-  padding-top: 55%;
+  max-height: 210px;
```
davidyuk, line 42: @AtanasKrondev how height should be actually defined? I can't get it from the code

AtanasKrondev: According to the design it should be 16:9 ratio.

AtanasKrondev:
```suggestion
  aspect-ratio: 16/9;

  @supports not (aspect-ratio: 16/9) {
    max-height: 360px;
    @include smallest {
      max-height: 210px;
    }
  }
```

<a href="https://caniuse.com/mdn-css_properties_aspect-ratio" target="_blank">Compatibility</a>. Honestly I didn't know about the ```aspect-ratio``` property and I don't vouch for it but seems suitable in this case.🤷🏻‍♂️

Liubov-crypto (commented):
Tested in Safari, FF, Chrome browsers with extension and iframe wallet, LGTM:
Trending - pass 
Search - pass
Change profile/back pics - pass 
Change profile info - pass 
Cookies - pass 
Tip on the Feed - pass 
Retip on the feed - pass  
Post without tipping - pass 
Comment - pass 
Tip comment on the feed - pass 
Pin/Unpin - pass 
Claim - pass 
Report post - pass 
Voting - pass 
Vote - pass 
Revoke Vote - pass 
Delegate voting power - pass 
Create a new poll - pass 
Search - pass 
Meet - pass 
Updates - pass 
Faq- pass 

and bugs remain #1111 #1112 (but maybe this is the side of the wallet)

src/components/tipRecords/TipMedia.vue
davidyuk: also in the failback, we can define the proper height for each breakpoint, check the last commit

AtanasKrondev (approved):

PR #1120, opened 2021-07-01
testnet config on develop adjustment

thepiwo:


.env.tnet
```diff
@@ -2,8 +2,17 @@ VUE_APP_BACKEND_URL=https://testnet.superhero.aeternity.art
...
-VUE_APP_WORD_REGISTRY_ADDRESS=ct_2sHzycCWkiyfZkCMoprn3Q7U8SQh3FYXp6iB1RAFVJy1pNWZAY
+
+VUE_APP_WORD_REGISTRY_ADDRESS=ct_2vTMtHgJTxTpsQ8Tm3o6DmfDkhc3vM4DefV845Yp9FSDuS6YWB
+VUE_APP_BONDING_CURVE_18_DECIMALS_ADDRESS=ct_2e1ZjvVXVi5f2ZiTqiFgs29oN1r1dhhdv2irvnnkvkoGZBgMPf
+
 VUE_APP_WORDBAZAAR_ENABLED=true
+VUE_APP_LANDING_ENABLED=
+
+VUE_APP_IMGUR_API_CLIENT_ID=eafea0a779e4039
+VUE_APP_GIPHY_API_KEY=P16yBDlSeEfcrJfp1rwnamtEZmQHxHNM
```
davidyuk, line 18:
```suggestion
```
Actually, it shouldn't be needed to redefine variables from `.env`, because they are always loaded:
https://cli.vuejs.org/guide/mode-and-env.html#example-staging-mode

thepiwo: cool, didn't know

davidyuk (approved):

PR #1124, opened 2021-07-03
refactor: drop bootstrap

davidyuk:
closes #740 

@Liubov-crypto would be nice to ensure that it doesn't add any significant visual bugs in both mobile and desktop

PR #1127, opened 2021-07-06
fix(UserInfo): avoid horizontal scrolling on mobile

davidyuk:
fixes issue

> It is possible to move the wallet to the left in the mobile browser

found in https://github.com/aeternity/superhero-ui/pull/1124#issuecomment-874191320


Liubov-crypto (approved):
Tested on Android FF and Mi browsers, LGTM.

AtanasKrondev (approved):

Issue #1126, opened 2021-07-05
Broken preview of posts in my profile channel tab


Liubov-crypto:
My profile, channel tab, where posts are pinned, the preview of posts is broken:
![preview](https://user-images.githubusercontent.com/69896204/124500149-e9763280-ddc7-11eb-9706-91342f8a65d0.png)

same on master and develop branch. 

@kenodressel could you please take a look?

davidyuk:
I see the same issue:
![Screenshot 2021-07-07 at 22 57 56](https://user-images.githubusercontent.com/9007851/124820943-c6d64c00-df76-11eb-9822-75142ed64b56.png)
https://superhero.com/users/ak_gvxNbZf5CuxYVfcUFoKAP4geZatWaC2Yy4jpx5vZoCKank4Gc

the problem is that `linkPreview` is not present in response of [`/pin/<address>`](https://raendom-backend.z52da5wt.xyz/pin/ak_gvxNbZf5CuxYVfcUFoKAP4geZatWaC2Yy4jpx5vZoCKank4Gc) endpoint

Liubov-crypto:
Preview from the site https://9gag.com is not available on the feed:
![9 gag](https://user-images.githubusercontent.com/69896204/124928121-026d2680-e008-11eb-9161-4c0985dff414.png)

https://superhero.com/tip/3073_v1

PR #1129, opened 2021-07-06
Use plain link in TipComment and related refactorings

davidyuk:
Based on this approach:
https://css-tricks.com/nested-links/
https://heydon.github.io/Inclusive-Components/cards-pseudo-content-author-link/

Ideally, it would allow us to drop all `@click.stop` from components related to the feed.

Plain links have better accessibility, including navigation between pages in SSR with disabled js.

@Liubov-crypto ensure that feed and comments looks/works correctly.


AtanasKrondev (approved):

Liubov-crypto (commented):
Tested, in general LGTM. 

Shifted comment design for new wallet account:
new user 1
new user 2
the same on develop branch. 

also remains #1126

PR #1135, opened 2021-07-08
Refactor TipRecord and nested components

davidyuk:
@Liubov-crypto ensure that all kinds of tips look acceptable on desktop and mobile
I have collected all of them in my channel: https://feature-refactor-tip-record.aggregator.z52da5wt.xyz/users/ak_gvxNbZf5CuxYVfcUFoKAP4geZatWaC2Yy4jpx5vZoCKank4Gc/channel
fixes #873

PR #1138, opened 2021-07-08
refactor(TipRecord): use plain link

davidyuk:
Approach from #1129 applied to TipRecord based on #1135

The only significant change is that the preview always points to a link that this preview for instead of a link to tip comments. This looks reasonable and was easier to implement. @onvisions is it acceptable?

@Liubov-crypto ensure that all kinds of tips behave the same as before except for the issue mentioned above.

### DIFF
--- a/src/components/AlertModal.vue
+++ b/src/components/AlertModal.vue
@@ -1,7 +1,7 @@
 <template>
   <div
     class="modal-mask"
-    @click.stop="resolve"
+    @click="resolve"
   >
     <div class="modal-wrapper">
       <div class="success-modal">
--- a/src/components/CookiesDialog.vue
+++ b/src/components/CookiesDialog.vue
@@ -1,8 +1,5 @@
 <template>
-  <div
-    class="cookies-dialog"
-    @click.stop
-  >
+  <div class="cookies-dialog">
     <ButtonPlain
       class="button-cancel"
       @click="$emit('close', $event)"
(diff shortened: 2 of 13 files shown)


Liubov-crypto (commented):
Tested in Safari, FF, Brave, and mobile browsers (FF and Xiaomi), in general LGTM. 

Except for a little bug, I can slide the feed to the left in Xiaomi mob browser (ver. 12.9.4-go):
зображення_viber_2021-07-09_18-19-29-730

davidyuk:
@Liubov-crypto Isn't it a browser's gesture to navigate to the next page?

| If the last page | If navigated to previous page before |
|---|---|
| Screenshot 2021-07-13 at 04 45 15 | Screenshot 2021-07-13 at 04 44 15 |

AtanasKrondev (approved):
LGTM, I couldn't reproduce it on Mi Browser 12.11.4-gn neither as a bug nor as a navigation gesture.

Liubov-crypto:
> Isn't it a browser's gesture to navigate to the next page?

No, I only had one page open. I checked again and noticed that this behavior only applies to posts "Tip Url". 
зображення_viber_2021-07-13_12-43-55-913
зображення_viber_2021-07-13_12-43-56-234


Idk, maybe my browser is old, but I have updated to V12.10.5-go.

davidyuk:
1. I mixed up Mi Browser Pro and UC browser
1. I have reproduced and fixed a flexbox issue

| Before | After |
|---|---|
| Screenshot 2021-07-13 at 21 47 44 | Screenshot 2021-07-13 at 21 57 56 |

Still preview image height is too high, I will check it

davidyuk:
🎉

PR #1139, opened 2021-07-09
Show several popups using popperjs

davidyuk:
depends on #1138 

Extracting into separate components:
- UserPopup (Popperjs that used in Bootstrap makes it drop up if no space at the bottom, see screenshot)
- FeedItemMenu (extracts some logic from TipRecord)
- TokenSelect popup (makes to use the same component in 3 places, @onvisions is it a design requirement for them to be different?)
- TipInputPopup (extracts some logic from TipInput that is not used in deeplink use case)
- CookiesDialog

Screenshot 2021-07-09 at 21 24 51

@Liubov-crypto please ensure that mentioned components work correctly


AtanasKrondev:
There's some weird flicker on the cookies popup while you are scrolling.

flick

davidyuk:
@AtanasKrondev I've fixed it by disabling a popper plugin that is enabled by default

AtanasKrondev (approved):
Nice!

Liubov-crypto (commented):
Tested, in general LGTM.

davidyuk:
> The wallet seems to have a broken dropdown (scroll doesn't work), I see fewer tokens than the dropdown below the post

cc: @etharner

davidyuk:
I've fixed TokenSelector by adding space between token name and amount

PR #1179, opened 2021-11-03
ci(docker): add gh action workflows

mmpetarpeshev:
The PR includes the following workflows: 
1.SSR docker pipeline for docker image
2.WEB docker pipeline for docker image
3.Staging deployments
4.Remove dockerhub tags on undeploy

Please take a look the on events , so we can confirm or change them by the requirements.
Once we have staging eks cluster , will remove the filter for docker images push to dockerhub.
For now we don't want to push images , while we don't have cluster for deploy.

**Do we want to deploy pr's on master ? Is the github workflow accepted that we merge from develop -> master and we don't want to repeat builds/deployments ?**

What left to be done , so we can use all workflows is : 
1. Paste comment in PR with the deployed url. - done
2. Make ArgoCD to work with staging branch.  - done
3. Deploy staging eks cluster.  - done
4. Test everything from end to end. - Only the production deployments events should be decided.
5. Make docker tags on PR to be the branch , so it can fit in the deploy workflow. - done
 
 **Production deployments needs change in composite gh actions and prod eks cluster available, we must not use them for now!**


kenodressel:
> Do we want to deploy pr's on master ? 

Yes

> Is the github workflow accepted that we merge from develop -> master and we don't want to repeat builds/deployments ?

We do develop -> release branch -> master and we want a deployment for that. I would say we could skip the pr from master to develop but i guess thats not worth the hassle.

> Test everything from end to end. - Here we hit a gh actions limitation for the workflow_run functionality.
To trigger a workflow runs on the workflow_run event, the workflow file should exist on the master or default branch.
We want to used the workflow_run event , because we want to trigger staging deploy only if the web or ssr docker build is successful.!! - Please comment how we want to proceed here , firstly merge and then check that all works good.

I guess we can do a formal review and see if the pipelines all work correctly. Then we merge it and if it does not work we fix it up afterwards.

kenodressel:
> we need to clear and what tags we want.

Could you elaborate about this a bit? You mean which tags we want to have permanently?

.github/workflows/ssr-docker.yml
```diff
@@ -0,0 +1,113 @@
...
+          startsWith(github.ref, 'refs/tags/v') ||
+          github.event_name == 'pull_request' && github.event.action == 'opened'
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: aeternity/superhero-ui-ssr
+          tags: |
+            type=raw,value=latest,enable=${{ endsWith(GitHub.ref, 'main') }}
+            type=ref,event=tag
+            type=ref,event=pr
```
kenodressel, line 51: this looks to me like all tags are applied to the same image? is that intentional?

.github/workflows/ssr-docker.yml
```diff
@@ -0,0 +1,113 @@
...
+    tags:
+      - 'v*'
+  pull_request:
+    branches: [ master, develop ]
+    types: [ opened, closed ]
+
+env:
+  PR_NUMBER: ${{ github.event.pull_request.number }}
+  ENV: "stg"
+  argocd-file: "values-staging.yaml"
```
kenodressel, line 16: should be `ARGOCD_FILE`

.github/workflows/web-docker.yml
```diff
@@ -0,0 +1,119 @@
...
+        with:
+          ref: develop
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 100
+
+      - if: github.ref != 'refs/heads/develop'
+        run: git branch develop origin/develop
+
+      - uses: actions/cache@v2
```
kenodressel, line 34: why do we cache here and not in the ssr?

.github/workflows/ssr-docker.yml
```diff
@@ -0,0 +1,113 @@
...
+      - uses: actions/checkout@v2
+        with:
+          repository: aeternity/gitops-apps.git
+          ref: staging
+          persist-credentials: false
+          fetch-depth: 0
+
+      - name: Install yq
+        run: |
+          sudo snap install yq
```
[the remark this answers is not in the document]
mmpetarpeshev: I moved it inside the composite action , will remove this step.We need yq to modify application helm chart values files, which are monitored and deployed by argocd

.github/workflows/ssr-docker.yml
```diff
@@ -0,0 +1,113 @@
...
+      
+      - name: Build and push docker image
+        if: |
+          github.event_name == 'push' ||
+          startsWith(github.ref, 'refs/tags/v') ||
+          github.event_name == 'pull_request' && github.event.action == 'opened'
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: Dockerfile.web
```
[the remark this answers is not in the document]
mmpetarpeshev: Correct , thanks. Will fix

.github/workflows/ssr-docker.yml
mmpetarpeshev: They are applied depends on the event

.github/workflows/ssr-docker.yml
mmpetarpeshev: Yep. will fix

.github/workflows/web-docker.yml
mmpetarpeshev: Let me check old workflows , maybe is left over from what I copied from the old workflows

.github/workflows/ssr-docker.yml
```diff
@@ -0,0 +1,114 @@
...
+        with:
+          fetch-depth: 100
+
+      - if: github.ref != 'refs/heads/develop'
+        run: git branch develop origin/develop
+      
+      - uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-${{ hashFiles('**/package-lock.json') }}
```
kenodressel, line 36: why are we caching this if we do not do anything with the cache?

.github/workflows/ssr-docker.yml
```diff
@@ -0,0 +1,114 @@
...
+          github.event_name == 'push' ||
+          startsWith(github.ref, 'refs/tags/v') ||
+          github.event_name == 'pull_request' && github.event.action == 'opened'
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: Dockerfile.ssr
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
```
kenodressel, line 69: this seems to not be set, should we remove it?

.github/workflows/ssr-docker.yml
kenodressel: we should also check https://github.com/docker/metadata-action#flavor-input, it seems by default `latest` tag is set to any pushed image

mmpetarpeshev: Yes , do we want this ?

.github/workflows/ssr-docker.yml
mmpetarpeshev: I have the same question here .Do we need this cache at all ?

.github/workflows/ssr-docker.yml
mmpetarpeshev: Not sure , that I understand the question correctly?

.github/workflows/ssr-docker.yml
mmpetarpeshev: I can explicitly set it to false

.github/workflows/ssr-docker.yml
kenodressel: Well we can use it but then we should use it to skip some steps. I don't see a reason here though. So lets remove it for now.

.github/workflows/ssr-docker.yml
kenodressel: ah nvm, its generated by the docker-meta action right?

.github/workflows/ssr-docker.yml
mmpetarpeshev: Sure , will remove it for now

.github/workflows/ssr-docker.yml
mmpetarpeshev: Yes

Issue #1193, opened 2021-11-24
Don't link to aeternal.io


davidyuk:
- open https://superhero.com/users/ak_z7HY9ApRm79oLqdomkvJ11r1KC4BY4APHaLYFZSTbW598EzvE
- press chain name at the top (`metaaeternity.chain`)
- it will redirect to https://mainnet.aeternal.io/account/transactions/ak_z7HY9ApRm79oLqdomkvJ11r1KC4BY4APHaLYFZSTbW598EzvE that is not working

PR #1245, opened 2024-04-29
refactor: upgrades sdk dependency to 13.3.2

kenodressel:


src/store/modules/aeternity.js
```diff
@@ -114,95 +97,125 @@ export default {
...
+          if (!newWallet) return null;
           clearInterval(webWalletTimeout);
           await sdk.connectToWallet(await newWallet.getConnection());
-          await sdk.subscribeAddress('subscribe', 'current');
-          const address = sdk.rpcClient.getCurrentAccount();
-          if (!address) return;
-          detector.stopScan();
+          const ret = await sdk.subscribeAddress('subscribe', 'current');
+          const { address: { current } } = ret;
+          const address = Object.keys(current)[0];
```
davidyuk, line 119:
```suggestion
          await sdk.subscribeAddress('subscribe', 'current');
          const address = sdk.addresses()[0];
```

src/store/modules/aeternity.js
```diff
@@ -114,95 +97,125 @@ export default {
...
       // eslint-disable-next-line no-underscore-dangle
       const webWalletTimeout = this._vm.$isMobileDevice ? 0
         : setTimeout(() => commit('enableIframeWallet'), 10000);
 
-      return new Promise((resolve) => {
-        detector.scan(async ({ newWallet }) => {
-          if (!newWallet) return;
+      async function handleNewWallet({ newWallet }) {
+        try {
+          if (!newWallet) return null;
```
davidyuk, line 114: this check seems to be not necessary
https://github.com/aeternity/aepp-sdk-js/blob/2743e4af828902d90e532447b6c526c833869e49/src/aepp-wallet-communication/wallet-detector.ts#L34-L50

src/store/modules/aeternity.js
```diff
@@ -114,95 +97,125 @@ export default {
...
+          return address;
+        } catch (e) {
+          return rejected(e);
+        }
+      }
+      console.log('Scanning for wallets...');
+
+      const scannerConnection = new BrowserWindowMessageConnection({
+        connectionInfo: { id: 'spy' },
+      });
```
davidyuk, line 133:
```suggestion
      const scannerConnection = new BrowserWindowMessageConnection();
```
https://github.com/aeternity/aepp-sdk-js/blob/2743e4af828902d90e532447b6c526c833869e49/examples/browser/aepp/src/Connect.vue#L89-L91

src/store/modules/aeternity.js
```diff
@@ -63,45 +52,39 @@ export default {
     },
   },
   actions: {
-    async initTippingContractIfNeeded({
-      commit,
+    async initTippingContract({
```
davidyuk, line 55: you can keep `IfNeeded` behaviour by putting contracts to a variable outside state, anyway contracts not expected to be reactive 🤔

vue.config.js
```diff
@@ -110,6 +110,36 @@ module.exports = {
...
+            '@babel/plugin-transform-class-properties',
+            '@babel/plugin-transform-logical-assignment-operators',
+          ],
+          ...defaultJsOptions,
+        },
+      },
+     */
+    config.module
+      .rule('babel-loader-for-apidom')
+      .test(/@swagger-api\/apidom-.*\.[mc]?js$/)
```
davidyuk, line 130: I think "swagger-client" was used only in sdk@8, is it still needed?

davidyuk (approved):
In general looks good

package.json
```diff
@@ -43,7 +43,6 @@
     "fontsource-ibm-plex-sans": "^3.1.5",
     "i18n": "^0.13.3",
     "is-fqdn": "^2.0.1",
-    "jitsi-iframe-api": "^1.0.0",
```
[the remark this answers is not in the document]
kenodressel: I am not sure where it comes from. I think its similar to the official jitsi package https://www.npmjs.com/package/@jitsi/web-sdk which also has been deprecated already. Since the servers are not running at the moment anways, I don't see a good reason to pour in the extra work to get it up and running.

src/store/modules/aeternity.js
kenodressel: How expensive is it to initialize the contracts when we need them? At the moment we only need them to do stateful actions (eg. post a tip or similar). And when we have another 10ms then I don't mind creating a fresh instance.

vue.config.js
kenodressel: Unfortunately many of the older repos (eg. BondingCurve) still require sdk 8 or similar. Therefore having this allows us to still use these dependencies. :cry:

package.json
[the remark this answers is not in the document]
davidyuk: I didn't know that servers also not running 😆

src/store/modules/aeternity.js
davidyuk: Not expensive at all because don't need to compile if ACI provided 👌

vue.config.js
kenodressel: Removed :)

thepiwo (approved):

CedrikNikita:
Shall we maybe include fix of that in the next release?
https://github.com/aeternity/superhero-ui/issues/1193

